PR #9, opened 2016-12-06
Bash tags

Spu00Key:
Large group of commits to re add the Skyrim Bash Tag Lists (including the ones Wyre Bash does not use).  They are useful in other patch merges and in aiding load order conflict resolution.

Also updated the bash tags for the group of addons I currently use/testing.


Ortham:
Have you checked if all these are actually useful or necessary? For example, `Crafting Materials.esp` has `{{BASH:C.Regions,C.Water}}` suggested, but I don't see how cell regions or water have anything to do with crafting materials. Not all possible tags are useful tags.

Spu00Key:
This may well be a learning opportunity for me; I noticed a lot of what you are saying with the ????.

Using your example; we get a C.Water tag when it comes to Riverwood.  Crafting Materials reverts a change made in Update.esm to XCWT - Water (Water Flow) back to what it was in Skyrim.esm.  The Update.esm either renamed or removed the water flow direction.

Why I consider this important?  The Water Mods (like Water 2 that I use) desire to be placed after anything that edits water flows.  Crafting Materials edits 4 areas in Riverwood when it comes to Water Flow data.  In all cases it overwrites the changes made by the Update/DLC ESMs and changes it back to the Skyrim defaults.

I'll take a leap of faith and say that the Water Mods change a lot of the water data away from the Skyrim defaults; and as a result, I may want to look deeper into the load order between the Water Mods and Crafting Materials.  Since really, I get the ??? reaction you got to Crafting Mods getting a C.Water tag; I'm likely to load it before any Water Mods.

There was another distinct example with a mod adding Lydia (and more) face presets to character generation.  It got tagged as modifying Actor data...  Again, anything else that modifies Actor Data will require a closer look at load order.   Now.... this particular mod should not be editing anything Actor Data; doing some tests it also contributes to a Lydia Grey Face Bug.  I've contacted the modder to let them know / check if the changes are intentional.

(Second Comment to break Wall of Text Next)

Spu00Key:
> Not all possible tags are useful tags.
This may be a big learning op or difference in opinion.  When I make these changes (either with loot or adding it to the ESP header data) and then export the tags from Wyre Bash... it doesn't care about any of the edits unless they mess with Inventory or Leveled Lists.

I think this is likely why the master file did not even list the extra tags as options for auto complete (even though several well established mods by Arthmoor have these tags in the header data).
Adding the data for me is more for visibility in the list rather than a component of another automatic function (although I think 1 other merger uses the extra tags).

When I see 7 mods of mine with the Stats tag on them it makes me want to look further at the changes and determine if load order should be adjusted. Maybe embarrassingly enough, I've caught 3 mods that modify 2 overlapping stats each, making one of them unnecessary.

Spu00Key:
So, I would like to close this pull request and make a new one.  The diff view is very hard to read now.  I'll restructure it and resubmit. 

Yes?

Ortham:
> Using your example; we get a C.Water tag when it comes to Riverwood. Crafting Materials reverts a change made in Update.esm to XCWT - Water (Water Flow) back to what it was in Skyrim.esm. The Update.esm either renamed or removed the water flow direction.
> 
> Why I consider this important? The Water Mods (like Water 2 that I use) desire to be placed after anything that edits water flows. Crafting Materials edits 4 areas in Riverwood when it comes to Water Flow data. In all cases it overwrites the changes made by the Update/DLC ESMs and changes it back to the Skyrim defaults.

In this case, water edits have nothing to do with crafting, so you'd never want to keep the Crafting Materials water edits (which are essentially wild edits), so there's no point suggesting C.Water for it. On the other hand, making water edits is the point of a water mod, so its C.Water suggestion is one you'd want.

> So, I would like to close this pull request and make a new one. The diff view is very hard to read now. I'll restructure it and resubmit.

Sure, but could you also check the tags you're suggesting: as above, if a tag is listed for a mod, but that tag is for something that isn't relevant to what the mod does, don't add it as a suggestion. Otherwise users might see the tag, assume that means the changes are important, and end up with a non-ideal situation.

Spu00Key:
Ahh, gotcha.  I have the thought backwards.  I was going under the if you see the tag then you need to check load order vs the intended if you see the tag then it's important.

TY

Ortham:
No, Bash Tags are used to partially ignore load order. For example, if mod A changes item weights and names, and mod B changes item weights and values, and for whatever reason B must load after A but you want A's names and B's weights, you can tag A with the Names tag and import it into your Bashed Patch, so the Bashed Patch will override B's names with the imported names from A.

Issue #8, opened 2016-12-03
Realistic Lighting Overhaul SSE


pStyl3:
[Realistic Lighting Overhaul SSE](http://www.nexusmods.com/skyrimspecialedition/mods/844)

We should ensure that the plugins from this mod will load in the following order (as laid out [here](http://www.nexusmods.com/skyrimspecialedition/articles/44/?)):

> Realistic Lighting Overhaul - Major City Interiors.esp
Realistic Lighting Overhaul - Minor Cities and Town Interiors.esp
Realistic Lighting Overhaul - Major City Exteriors.esp
Realistic Lighting Overhaul - Dawnguard Interiors.esp
Realistic Lighting Overhaul - Dungeons.esp
Realistic Lighting Overhaul - Weathers.esp
Realistic Lighting Overhaul - Dawnguard Weathers.esp
RLO - Major City Exteriors - No Guard Torches.esp

PR #10, opened 2016-12-07
Issue #8 realistic lighting overhaul sse

Spu00Key:
#8 Made requested changes.

Is there a more, elegant way to do this?

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -101,6 +101,10 @@ plugins:
   - name: 'TouringCarriages.esp'
     global_priority: -90
   - name: 'Cutting Room Floor.esp'
+    global_priority: -90
+    after:
+      - 'TouringCarriages.esp'
+      - 'WhiterunHoldForest.esp'
     tag:
       - C.Location
       - C.Owner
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -113,7 +113,49 @@ plugins:
       - Graphics
       - Invent
       - Names
-      - Stats
+      - Stats
```
Ortham: Trailing whitespace!

Ortham (changes requested):
No, that's how you do it. What's up with the merge commit though? You could just have done a fast-forward merge, or no merge at all, both would be cleaner.

I may be able to squash that commit out when I merge, but please fix the trailing whitespace.

Spu00Key:
Awesome; sorry about the trailing spaces. Still learning on the Git Options.

So I looked up the [fast forward merge](https://ariya.io/2013/09/fast-forward-git-merge).  

For this option I need to pull down the pull request and squash all commits; then the system should auto fast forward instead of a branching commit.

You also mentioned doing a no merge.  I've not found a way to do that through my internet searching.

For my own learning, I'm going to close this and try to get the fast forward working.

PR #11, opened 2016-12-07
Fix for Issue #8

Spu00Key:
Adds sort order for Realistic Lighting Overhaul modules.

Also, attempting a fast forward as discussed in #10.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -114,6 +114,48 @@ plugins:
       - Invent
       - Names
       - Stats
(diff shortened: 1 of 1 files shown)


Spu00Key:
Nope on #10.  Not having write access is where I think it is breaking.  When doing this within my own fork, "This branch has no conflicts with the base branch" has a merge/pulldown option.

Here it just says:
"Only those with write access to this repository can merge pull requests"

PR #13, opened 2016-12-15
Updates to load order, added notes and warnings.  Added RDO

Spu00Key:
- Elysium Estate warnings on modifiers.
- Verdent load order.
- Bijin notes on ESP active (is there a way to check for this and
display a warning instead of a note?)
- Laterns wants to be before USSEP
- Added Relationship Dialogue Overhaul and a few of its many load order
needs.


Ortham:
What's the point of having those plugins installed if they're supposed to be inactive, and why shouldn't they be active? Would a "do not use this mod" message not be more appropriate?

Spu00Key:
A lazy author...  He is using Skyrim assets for a SSE mod.

The author has an all in one ESP for SE for all his Bijin NPCs in Skyrim.
- So you go over to Skyrim, download 5 mods
- Come back here and download the SE combiner mod.
- Install everything, but deactivate the Skyrim Bijin or you CTD 

Everything works fine this way.  I was wondering if I could show a note or warning only if the esp was enabled?

Ortham:
> Everything works fine this way. I was wondering if I could show a note or warning only if the esp was enabled?

Yes, the syntax for doing that is:

```
    name: 'Foo.esp'
    msg:
      - type: say
        content: 'Disable ESP'
        condition: 'active("Foo.esp")'
```

Spu00Key:
Busy week.  My changes are complete.  Anything to fix?

Ortham:
Not that I can see.

pStyl3:
> - name: 'Eli_Breezehome.esp'
> after:
>- 'Immersive Citizens - AI Overhaul.esp'

According to the mod description of [Immersive Citizens - AI Overhaul](http://www.nexusmods.com/skyrimspecialedition/mods/173), `Immersive Citizens - AI Overhaul.esp` must load after `Eli_Breezehome.esp`.

So this would need to be switched around.

Ortham:
> So this would need to be switched around.

Fixed in 410e3d9ae21c20e99333990436ce414933044d85, thanks.

PR #18, opened 2017-01-01
MaddPatch

Spu00Key:
- Several Leveled List type mods did not have their bash tags set.
- - Verified through xEdit

- Added the birds (all of them) and ducks.
- - Ducks want to swim to the top of the list
- - The Birds match up a certain load order

- Two town mods wanted to be after IC

- Bring out the dead wanted to be at the top of the list.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -355,6 +355,7 @@ plugins:
   - name: 'Whistling Mine.esp'
     after:
       - 'Verdant - A Skyrim Grass Plugin SSE Version.esp'
+      - 'Immersive Citizens - AI Overhaul.esp'
   - name: 'RealisticWaterTwo.esp'
     after:
       - 'JK''s Riverwood.esp'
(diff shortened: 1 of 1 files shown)


Spu00Key:
Pulled the trigger too soon on that one. My local testing broke somewhere.

masterlist.yaml
```diff
@@ -578,3 +581,50 @@ plugins:
...
+      - Delev
+      - Relev
+  - name: 'CraftingSupplies_AIO_ALL.esp'
+    tag:
+      - Invent
+  - name: 'Ducks and Swans.esp'
+    global_priority: -90
+  - name: 'BirdsOfSkyrim_SSE.esp'
+    after:
+      - 'Skytest'
```
Ortham: This is missing a file extension.

Ortham (changes requested):

masterlist.yaml
Spu00Key: Fixed!

Spu00Key:
Should I add a -1 Global to mods that directly state they can be loaded anywhere in the list?

Ortham (approved):

Ortham:
> Should I add a -1 Global to mods that directly state they can be loaded anywhere in the list?

No, -1 would mean they should be loaded before any mods which don't have a lower priority.

Spu00Key:
That is the desired effect.

Ortham:
"Can load anywhere" is very different from "must load before most plugins".

Spu00Key:
Awesome, ty.  I appreciate all the help you've been giving me.  I've got about half of the documents read; but will likely go back through the first parts again now that I'm dangerously familiar with the process.

Issue #25, opened 2017-03-06
Problem with FlowerGirls and Relatiopnship Dialog Overhaul order


PauloMBrito:
Latest update 0.10.03 /
Masterlist Revision a688e66
Masterlist Date 2017-03-04
Don't take in consideration that Flower Girls must be After Relationship Dialogue Overhaul.

Author of Flower Girls (Xiderpunk) say
Flower Girls (FG) alters some lines of dialogue which are also effected by Relationship Dialogue Overhaul (RDO). Please ensure Flower Girls is AFTER RDO in your load order.

NMM 0.63.13 / win10pro x64

In mine configuration LOOT put FG in position 59 and RDO in 123. 

[CEFDebugLog.txt](https://github.com/loot/skyrimse/files/822878/CEFDebugLog.txt)
[LOOTDebugLog.txt](https://github.com/loot/skyrimse/files/822877/LOOTDebugLog.txt)

Spu00Key:
I'm not reading into the two logs attached as that is out of my scope.

Both mod Author's have been contacted to verify issue.
X Edit confirms 7 topic conflicts, although I don't see where they would cause any issue.

@PauloMBrito what in game issues have you had with the incorrect load order?

PauloMBrito:
So far none that i found, but i send this message / open this issue because see the warning and detected that LOOT don't do that.
PS - Because i don't know what lines are affected, i can't know if that conflict exist.

Spu00Key:
Ok; if I get word back from one of the Author's I'll make the change.

The conflicting topics are when you are selling harvest items to someone:
- I have cabbages to sell

And so on.

PauloMBrito:
:)

Spu00Key:
@WrinklyNinja, PR #27 closes this issue.

PR #27, opened 2017-03-11
Adds FlowerGirls and places it after RDO

Spu00Key:
Resolves issue #25


Spu00Key:
Author of Flower Girls verified change.

PR #33, opened 2017-06-24
Added support for my own mod: EBM_TrainMore

ExE-Boss:
My mod EBM_TrainMore is available in multiple versions, which are incompatible with each other.
I’ve added that information into the masterlist.

### External links:
- [http://www.nexusmods.com/skyrimspecialedition/mods/10580](http://www.nexusmods.com/skyrimspecialedition/mods/10580 "ExE Boss's Mods - Train More at Skyrim Special Edition Nexus - Mods and Community")


Ortham:
Thanks!

PR #44, opened 2017-09-06
Add ULALPatch dirty info

BeermotorWB:
Added xEdit dirty plugin info for ULALPatch.esp from Unique Locations - Riverwood Forest.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -706,3 +706,11 @@ plugins:
       - 'Verdant - A Skyrim Grass Plugin SSE Version.esp'
   - name: 'DynDOLOD.esp'
     global_priority: 115
+  - name: 'ULALPatch.esp'
+    after:
+      - 'Alternate Start - Live Another Life.esp'
+    dirty:
+      - crc: 0xdde49e20
+        <<: *dirtyPlugin
+        itm: 4
+        udr: 1


Ortham:
Looks good!

Sharlikran:
patchheader

@BeermotorWB One of the cool things about LOOT is that it automatically sorts plugins after their respective masters. For example in the above screen shot you see that ULALPatch.esp has two masters, Alternate Start - Live Another Life.esp  and UniqueLocationsRiverwoodForest.esp.

This would be incorrect:
```
  - name: 'ULALPatch.esp'
    after:
      - 'Alternate Start - Live Another Life.esp'
      - 'UniqueLocationsRiverwoodForest.esp'
    dirty:
      - crc: 0xdde49e20
        <<: *dirtyPlugin
        itm: 4
        udr: 1 
```
This would be correct:
```
  - name: 'ULALPatch.esp'
    dirty:
      - crc: 0x6d3163ad
        <<: *dirtyPlugin
        itm: 4
        udr: 1 
```

It was a quick fix so I took care of it. Good job so far. :+1: 

One use of after metadata is for plugins that don't share the same master but edit the same records. If a conflict exists you would examine the plugins in xEdit and determine if order will resolve the conflict. If order will resolve the conflict and still maintain the authors intended changes, then you would add after metadata. If order will not resolve the conflicts and doesn't maintain the authors intended changes, then they would be incompatible.

BeermotorWB:
Splendid and thank you.  This was good practice and I feel a lot more solid
about editing now.

BeermotorWB:
Also @Sharlikran  that makes perfect sense. Thanks!

PR #45, opened 2017-09-10
Bash tag adds

BeermotorWB:
Added bash tags for several extant and new plugins.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -406,7 +406,6 @@ plugins:
       - 'SoS - Civilization.esp'
     tag:
       - C.Location
-      - C.Owner
       - C.Regions
       - C.Water
       - Graphics
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -349,6 +349,9 @@ plugins:
   - name: 'JKs Whiterun.esp'
     after:
       - 'Verdant - A Skyrim Grass Plugin SSE Version.esp'
+    tag:
+      - C.Location
+      - C.Water
   - name: 'JKs Whiterun exterior.esp'
     after:
       - 'Verdant - A Skyrim Grass Plugin SSE Version.esp'
@@ -401,6 +404,12 @@ plugins:
       - 'SoS - The Wilds.esp'
       - 'SoS - The Dungeons.esp'
       - 'SoS - Civilization.esp'
+    tag:
+      - C.Location
+      - C.Owner
+      - C.Regions
+      - C.Water
+      - Graphics
   - name: 'SSEMerged.esp'
     global_priority: 80
     tag:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -712,3 +721,67 @@ plugins:
         <<: *dirtyPlugin
         itm: 4
         udr: 1
+  - name: 'AyleidCitadel.esp'
```
Freso, line 723: It would be nice to have the `url` included for new entries as well, allowing other/future editors to easily look up the plugin. :) (Also, I've been talking a bit with [modwat.ch](https://modwat.ch/) about them using the LOOT masterlists to add links to plugins in their lists, which would also need the `url` attributes being set.)

masterlist.yaml
```diff
@@ -712,3 +721,67 @@ plugins:
...
+  - name: 'DBM_CRF_Patch.esp'
+    tag:
+      - Factions
+      - Names
+  - name: 'DBM_SkyrimUnderground_Patch.esp'
+    tag:
+      - Invent
+  - name: 'DBM_HelgenReborn_Patch.esp'
+    tag:
+      - Factions
```
Freso: If you can set whatever editor you use to add a new line/line break at the end of the file, that would be great. Not really a big deal, but it prevents commits "fighting" over changing the last line as some editors add it back and others remove it and just clutters diff views unnecessarily.

Freso (commented):
Overall LGTM, but I'll leave it for someone more familiar with SSE to merge. :)

masterlist.yaml
BeermotorWB: I took it out on purpose actually since I'm used to dealing with yaml on non-Windows platform,  but I'll try to retain the cr/lf at the end from here on.

Freso: (Note that the line breaks will be converted by git when cloning the repository, so while it's `CRLF` for you, it's `LF` for me (on Linux) and will be just `CR` for people on Mac OS <X machines. :) Git does the line ending conversion transparently (unless you explicitly tell it not to!).)

BeermotorWB: Good to know.  In my day job I'm a *nix admin but our desktop machines are all MBPs, so I've been insulated from the Windows CRLF mess for the past decade or so. 

On the Windows gaming box I have at home I'm using Notepad++ which should respect the existing format of the file, but I'm thinking I'll just set it to use LF for YAML files to play it safe. That or I could just use Vim in a WSL terminal like a civilized human.  ;)

BeermotorWB:
I think I was on autopilot and forgot about the "delete your fork instead of syncing" recommendation in the documentation. 

I like the branch idea but I didn't want to presume I'd ascended to that branch status yet after only two PRs ;)

masterlist.yaml
```diff
@@ -401,6 +404,12 @@ plugins:
       - 'SoS - The Wilds.esp'
       - 'SoS - The Dungeons.esp'
       - 'SoS - Civilization.esp'
+    tag:
+      - C.Location
+      - C.Owner
+      - C.Regions
+      - C.Water
+      - Graphics
```
Ortham: I haven't looked at Realistic Water Two in a while, but are C.Location and C.Owner really appropriate?

Ortham:
My comments would echo Freso's, a trailing newline is standard practice, and providing the `url:` entries for new plugins can be useful (but I myself don't always do so). 

One branch per PR is also generally good practice, and you can then keep the branches from upstream (e.g. `v0.10`) in sync while branching off them to make your PRs, and keep history clean. In this case I can squash and merge, but I prefer PRs to be merged with merge commits, it makes it clear that the source was external.

Ortham:
We should probably update the contributing guide to mention this stuff...

masterlist.yaml
BeermotorWB: Location yes, but now that you mention it I want to look at C. Owner.

BeermotorWB: Actually @WrinklyNinja since Irma decided she wanted to visit Space Camp after visiting Disney World, I may not get to look at this tonight, especially if we don't have power.

I've edited my fork so I'm going to close this PR and submit a fresh one.

BeermotorWB:
~~Since Irma decided she wanted to go to Space Camp after visiting Disney World, I may not get to look at the realistic water two plugin tonight.~ ~ Irma spared Space Camp and our trees so I was able to get this done. 

I've made a couple of quick changes to my fork (I didn't realize they'd be inherited by this PR) to remove the tag in question and to add a proper LF at the EOF.

masterlist.yaml
BeermotorWB: OK it's raining sideways here but I figured out the C.Owner tag on RWT: the plugin sets player ownership of a RWT-placed horse trough (RWT_HorseTrough) in the various Hearthfires homes. 

It'll probably be fine to leave it out. If we ever hear of anything clobbering the RWT Horse Troughs we can add it back.

Ortham:
👍

PR #54, opened 2017-12-04
Update masterlist.yaml

Graybark:
Added Skyrim Hair Replacer Mod

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -370,6 +370,12 @@ plugins:
   - name: 'BetterQuestObjectives.esp'
     tag:
       - NoMerge
+  - name: 'Hair Replacer.esp'
+    global_priority: -100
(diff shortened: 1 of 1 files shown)


Ortham:
Do the global priorities really need to be -100? That's a very negative value, it doesn't leave much room for anything else...

Graybark:
Well, Cutting Room Floor has -90 and it should preferably load before it because I think that black faces are worse than old hair. This mod replaces pretty much all of the vanilla hair and if another mod messes with the facegen or headparts, this should load first for the same reason. All archived mods should load after it since this is a base replacer. I could also list every single mod that does that but the list would be incredibly long. Half of Arthmoors mods alone add facegen. What do you think?

Ortham:
Ok, that seems fine then.

Issue #56, opened 2018-01-15
Load RealisticWaterTwo.esp after Alternate Start - Live Another Life.esp and CFTO.esp


Riwaha:
It appears that NAM3 - LOD Water Type is overwritten by Alternate Start - Live Another Life.esp, and CFTO.esp.
It causes water seams to occur similar to ![Water Seams](https://i.imgur.com/GH5LBUm.png)
By placing RealisticWaterTwo.esp after those two mods, it then fixes it completely.
I suppose that CFTO is not as important, however Alternate Start is a very common mod.
The two mods are
[Alternate Start - Live Another Life](https://www.nexusmods.com/skyrimspecialedition/mods/272)
[Carriage and Ferry Travel Overhaul](https://www.nexusmods.com/skyrimspecialedition/mods/8379)

PR #57, opened 2018-01-16
Update Realistic Water Two load order

Riwaha:
Fixes water seam issues caused by Alternate Start - Live Another Life and Carriage and Ferry Travel Overhaul modifying cells near bodies of water.

Realistic Water Two states "This mod must be placed AFTER any mod that modifies a cell near a body of water!
This must be done to avoid seeing "seams" in the water. It is recommend placing this mod VERY low in the load order (towards the bottom) and use xEdit to make any necessary patches with other mods."

Fixes issue #56


Ortham:
Thanks 👍 Looks like Travis was having a problem with the build for a while, but it passed in the end.

PR #64, opened 2018-04-02
Update masterlist - SKSE 64 Version Check & Warnings

MacSplody:
Added version check for SKSE64 & one warning message for Skyrim Skill uncapper.

Edit: Will try implementing the check for exe versions and SKSE64 loader versions as in Fallout 4 masterlist  as Arthmoor suggested tomorrow.

Also are the groups naming scheme from Skyrim going to be carried over to SSE?


Ortham:
> Also are the groups naming scheme from Skyrim going to be carried over to SSE?

I haven't got any further changes to group naming planned. If it makes sense to, they can be, but my priority is to have groups working well in the first place.

Issue #48, opened 2017-10-01
Add entries for SKSE64 and SkyUI


pStyl3:
LOOT's [masterlist for Skyrim](https://raw.githubusercontent.com/loot/skyrim/v0.10/masterlist.yaml) contains some data for SKSE and SkyUI (at the very top of it).

Should we do the same for SSE's masterlist? [SKSE64](http://skse.silverlock.org/) is in alpha now, as is [SkyUI](https://www.nexusmods.com/skyrimspecialedition/mods/12604).

Or maybe wait until they reach a stable release?

CC: @schlangster @Mardoxx

Mardoxx:
Absolutely no idea. Ask maintainers of loot

Ortham:
I think it makes sense to add it when it's needed. At the moment, I don't think there is any need.

Arthmoor:
Does it make sense now that 1.5.3 isn't in beta anymore?

Ortham:
> Does it make sense now that 1.5.3 isn't in beta anymore?

That doesn't really change that SKSE64 and SkyUI are still alpha, though it does mean I should probably get on with ESL support for LOOT...

Arthmoor:
I don't think it would hurt to put the support in for testing if nothing else.

Ortham:
If it's needed for testing, then sure.

Arthmoor:
This probably needs to be looked at again to provide for exe versions and SKSE64 loader versions in the same way the Fallout 4 list currently is since both are likely to change again in the future.

MacSplody:
#65 Needs some verification, one issue with version check for SkyrimSE.exe, version listed in exe details is 1.0.0.0 not 1.5.23

Arthmoor:
The version number needs to be set to 1.5.39 now anyway since SSE just updated this morning.

MacSplody:
@Arthmoor  Unable to get the true version of the SKyrimSE.exe at the moment. Quoting @WrinklyNinja 

> Given that the executable version doesn't reflect the game version, I think it's pointless checking it. You could check the CRC instead, but I don't know if different regions get different executables (which would all have different CRCs)...

Arthmoor:
Not sure what to tell you then cause only Bethesda can control what shows up in the exe data.

pStyl3:
SKSE64 has left its alpha status and is now considered being in beta (v2.0.7).

MacSplody:
I'll close this issue as SkyUI & SKSE entries have been added. 
[...]

PR #65, opened 2018-04-03
[masterlist.yaml] SKSE Version checks & Messages

MacSplody:
- SKSE64 & SSE Exe version checks
- Compatibility check SkyrimSE.exe & skse64_loader.exe
- Message options for (RequiresSKSE, Obsolete, patched, redundant, conflicting & patch available) Mods.
- Warning message for missing SKSE loader when .dll requires it. (Could add catch all to look for any .dll in skse\plugin folder)


MacSplody:
One open issue is the SkyrimSE.exe version is currently listed as 1.0.0.0 in the exe details.

image

masterlist.yaml
```diff
@@ -112,9 +238,69 @@ globals:
...
       - lang: zh_CN
         text: '[最新的LOOT版本](%1%)。'
+      - lang: da
+        text: '[Seneste LOOT-tråd](%1%).'
       - lang: ja
         text: '[LOOTの最新スレッドを開く](%1%)（英語）'
     subs: [ 'https://loot.github.io/latest-thread/' ]
+# SKYRIM SE Latest
+  - type: say
+    condition: 'file("../SkyrimSE.exe") and version("../SkyrimSE.exe", "1.0.0.0", <)'
```
Ortham: Given that the executable version doesn't reflect the game version, I think it's pointless checking it. You could check the CRC instead, but I don't know if different regions get different executables (which would all have different CRCs)...

masterlist.yaml
```diff
@@ -112,9 +238,69 @@ globals:
...
+      - lang: ko
+        text: 'SKSE64가 최신버전입니다.'
+      - lang: zh_CN
+        text: '你的SKSE64是最新的。'
+      - lang: da
+        text: 'Din SKSE64 er opdateret.'
+      - lang: ja
+        text: 'SKSE64は最新版です。'
+  - type: error
+    condition: 'version("../skse64_loader.exe", "0.2.0.6", ==) and version("../SkyrimSE.exe", "1.0.0.0", !=)'
```
Ortham: Same as above, if the version is going to stay at 1.0.0.0, this isn't a terribly helpful message.

masterlist.yaml
```diff
@@ -112,9 +238,69 @@ globals:
...
+      - lang: en
+        text: 'Your version of skse64 is not compatible with your version of Skyrim Special Edition.'
+      - lang: da
+        text: 'Din version af skse64 er ikke kompatibel med din version af Skyrim Special Edition.'
+      - lang: ja
+        text: 'このバージョンのskse64は使用中のSkyrim Special Edition との互換性がありません。'
+# SKSE64 plugins requirement
+  - <<: *requireSKSE64
+    subs: [ 'Achievements Mods Enabler' ]
+    # url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/245' ]
```
Ortham: A more useful way of encoding the link would be to include it in the message substitution, e.g. `subs: [ '[Achievements Mods Enabler](https://www.nexusmods.com/skyrimspecialedition/mods/245)' ]`. That way it'll show up as a hyperlink in the UI and won't look slightly confusing - initially I wondered why you'd left invalid syntax commented out.

masterlist.yaml
```diff
@@ -19,8 +97,56 @@ common:
...
+  - &hasPluginPatch
+    type: warn
+    content:
+      - lang: en
+        text: "You seem to be using %1%, but you haven't enabled the %1% patch for this plugin: %2%"
+      - lang: da
+        text: "Du ser ud til at bruge %1%, men du bruger ikke “%1%”-lappen for dette plugin: %2%"
+      - lang: ja
+        text: "%1%が使用されていますが、このプラグイン用の%1%のパッチが有効になっていません: %2%"
+  - &requireSKSE64
```
Ortham: As far as I can see, this is the only `common` entry you've added that's actually used, why all the others?

Ortham (changes requested):
Thanks for this, I've left my comments inline. It's unfortunate that SkyrimSE.exe's file version doesn't reflect its real version.

masterlist.yaml
MacSplody: Yeah only noticed it late in testing, unusual as FO4 and Oblivion both display the correct version.

masterlist.yaml
MacSplody: Can't claim credit for the code, it was copied from FO4 masterlist. Thanks for the tip.

masterlist.yaml
MacSplody: Haven't used YAML before so was mainly doing it to get an idea of how it was working.
And also because those entries were in FO4 masterlist which I assumed was more active and up to date.

Ortham: OK, well the FO4 masterlist is the most recent, but if an entry in the `common` section isn't referenced by its `&anchor` anywhere, it's not used and so is just padding the masterlist with noise, so it's better for it not to be there.

Ortham:
If you remove the unused `common` entries and the messages that have a `SkyrimSE.exe` version check, I'll be happy to merge this.

MacSplody:
- Remove unused common entries.
- Remove SkyrimSE.exe version & SKSE compatibility check. (Unable to get true exe version at this time).
- Updated requireSKSE64 entry for Achievment mod with suggested change to link.

masterlist.yaml
```diff
@@ -112,9 +131,49 @@ globals:
...
       - lang: zh_CN
         text: '[最新的LOOT版本](%1%)。'
+      - lang: da
+        text: '[Seneste LOOT-tråd](%1%).'
       - lang: ja
         text: '[LOOTの最新スレッドを開く](%1%)（英語）'
     subs: [ 'https://loot.github.io/latest-thread/' ]
+# SKSE Latest
+  - type: say
+    condition: 'file("../skse64_loader.exe") and version("../skse64_loader.exe", "0.2.0.6", <)'
```
Ortham, line 141: I didn't think of this before, but this condition can be shortened to just `version("../skse64_loader.exe", "0.2.0.6", <)`, if the file doesn't exist it will be false.

MacSplody: I'll test it in a bit and push change.

MacSplody: Tested &  
- using either != and < would always show skse latest version message.
condition: 'version("../skse64_loader.exe", "0.2.0.6", !=)'
condition: 'version("../skse64_loader.exe", "0.2.0.6", <)'

Ortham: Ah, sorry, my mistake, `version()` will return true if the file doesn't exist for `!=`, `<` and `<=`, though thinking about it again I'm not sure what would be most intuitive.

Ortham:
Thanks 👍

MacSplody:
#48 Completed

PR #70, opened 2018-04-08
V0.10 masterlist updates

MacSplody:
Moon And Star
- Bash tags added
- Load after AuroraVillage.esp
- Dirty & clean plugin CRC & info.
- Add compatibilty info.

Undeath
- Load after Moon & Star, recommended by M&S author.
- Bash tags added
- Dirty & clean plugin CRC & info.
- Mod contains wild edits, link to manual cleaning guide provided.
- Add compatibilty info. AuroraVillage.esp is not compatible.

Improved Closeface Helmets
- Load after rule for Morrowloot ultimate.
- Compatibility warning for morrowloot & undeath, with links to patches provided.
- Bash tag (Graphics) suggested.
- Dirty & clean plugin info added.
- Link to patch for Improved closefaced helmets for Undeath SSE

Improved Closefaced Helmets Patches
- MLU patch
- Remove bad bash tag
- Flag required master, missing from file header.
- Add clean plugin info.

Morrowloot Ultimate
- Add suggested bash tags
- Add clean plugin CRC

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -466,8 +466,6 @@ plugins:
         udr: 0
         nav: 0
     clean:
-      - crc: 0x8249EC41
-        util: SSEEdit v3.2.1
       - crc: 0xB98044B4
         util: SSEEdit v3.2.1
 
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -426,6 +426,8 @@ plugins:
       - crc: 0x2BC2FECB
         util: SSEEdit v3.2.1
   - name: 'MoonAndStar_MAS.esp'
+    after:
+      - 'AuroraVillage.esp'
     tag:
       - C.Location
       - C.Water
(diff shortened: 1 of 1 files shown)


MacSplody:
Is it fine to add the translations with just the english like below?

        content:
          - lang: en
            text: 'This mod has certain conflicts with "MLU.esp". Consider using the [MLU - Improved Closefaced Helmets patch](https://www.nexusmods.com/skyrimspecialedition/mods/3058/).'

Ortham:
Yes, that's fine, it's just a bit more verbose.

masterlist.yaml
```diff
@@ -425,8 +425,118 @@ plugins:
...
+    tag:
+      - Delev
+      - Relev
+  - name: 'MLU - Improved Closefaced Helmets.esp'
+    # missing master.
+    req:
+      - 'MLU.esp'
+    tag:
+      - Stats
+      - --Names
```
Ortham: `--Names` looks like a typo - it should probably either be `Names` or `-Names` (the latter if you're suggesting the tag should be removed).

Ortham (commented):
I'm not sure why you moved the existing MLU entries, but it doesn't really matter. Aside from the suspicious Bash Tag suggestion, this looks good.

masterlist.yaml
MacSplody: It's how loot user meta produced the entry, I tried the suggested way first. -Names seems to add tags while --Names removes them. I expected it to work like so

         tag:
            - +Names  for adding
            -  -Names for removing
But instead it defaults to add.
               tag:
                 -  -Names  for adding or just - Names
                 -  --Names for removing

I only moved it for clarity at the time, but okay I'll avoid moving mods if that's an issue.

Ortham: That sounds like a bug... I'll take a look.

MacSplody:
tag:
      - --Stats
      - -Graphics

Is what is being generated currently for userlist btw. bottom one is add. Could be an issue with an extra - being added.

Ortham:
Yep, just tried it out and I've filed an issue for it.

PR #74, opened 2018-04-11
Master-list Update - Corrections

MacSplody:
Corrections & Updates for USSEP & Survival Patch
- Correct order of sections
- Add URLs and plug-in version information.


Ortham (commented):
Adding URLs is good, but there's no meaning to the order of keys. Why do you think they should be moved?

MacSplody:
Figured it should at least be consistent with other entries, makes it easier to see at a glance what keys are present. Will I revert since they don't matter?
I also have an indentation fix for bijin entries, if that's okay to add?

Ortham:
Consistency is good, but we don't currently have any linting for formatting, so it would be playing whack-a-mole. If you're going to make a lot of formatting changes, can you bundle them up into a separate PR from your actual metadata changes (in this case the URLs you've added)?

MacSplody:
Okay I can do that, so will I close this pull and make a separate pull with just metadata changes and another with format changes?

Edit: I'll just start a new pull with only metadata changes.

Ortham:
Yep, thanks.

PR #78, opened 2018-04-11
Fixed SkyrimIsWindy patch order

GonDragon:
Fixed an Issue where loot tried to load the "SkyrimIsWindy - EVT patch.esp" before the patched file "SkyrimIsWindy.esp".


Ortham:
Thanks 👍

PR #80, opened 2018-04-12
Metadata Update - BCS & BCS Lost Library - URL, Bash tags, Clean plug-in info & Load After Rules

MacSplody:
Book Covers Skyrim
- Load after rules taken from checking compatibility notes & conflicts in xEdit
- Unlimited Bookshelves, weightless books

BCS: Lost Library
- Delev tag, alters level list for merchants.

- Clean plugin info verified for both in xEdit

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1346,3 +1346,36 @@ plugins:
   - name: 'Convenient Horses.esp'
     after:
       - 'Open Cities Skyrim.esp'
+  - name: 'Book Covers Skyrim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/901/'
+        name: 'Book Covers Skyrim on Nexus Mods'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3071870/'
+        name: 'Book Covers Skyrim - Original on Bethesda.net'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3072082'
+        name: 'Book Covers Skyrim - Desaturated on Bethesda.net'
+    after:
+      - 'UnlimitedBookshelves.esp'
+      - 'Weightless Books.esp'
+      - 'WeightlessOV - Book.esp'
+    tag:
+      - Graphics
+      - Names
+    clean:
+      # version 4.2
+      - crc: 0x32587221
+        util: SSEEdit v3.2.1
+  - name: 'Book Covers Skyrim - Lost Library.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/902/'
+        name: 'Book Covers Skyrim on Nexus Mods'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3203152'
+        name: 'BCS Lost Library - Original on Bethesda.net'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3203032'
+        name: 'BCS Lost Library - Desaturated on Bethesda.net'
+    tag:
+      - Delev
+    clean:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1346,3 +1346,36 @@ plugins:
...
+        name: 'Book Covers Skyrim - Original on Bethesda.net'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3072082'
+        name: 'Book Covers Skyrim - Desaturated on Bethesda.net'
+    after:
+      - 'UnlimitedBookshelves.esp'
+      - 'Weightless Books.esp'
+      - 'WeightlessOV - Book.esp'
+    tag:
+      - Graphics
+      - Names
```
MacSplody, line 1363: Tags I got back from the script I'm using were {{BASH:C.Acoustic,C.Climate,C.Music,C.Owner,Graphics,Names}}

Manual checked them and only Graphics,Names are appropriate.

It does modify entries relating to C.Acoustic,C.Climate,C.Music,C.Owner but these are implicit changes forwarded from Unofficial Special Edition Skyrim Patch and there not relevant to the mod.

masterlist.yaml
```diff
@@ -1346,3 +1346,36 @@ plugins:
...
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/901/'
+        name: 'Book Covers Skyrim on Nexus Mods'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3071870/'
+        name: 'Book Covers Skyrim - Original on Bethesda.net'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3072082'
+        name: 'Book Covers Skyrim - Desaturated on Bethesda.net'
+    after:
+      - 'UnlimitedBookshelves.esp'
+      - 'Weightless Books.esp'
+      - 'WeightlessOV - Book.esp'
```
MacSplody, line 1360: Loot sorts these mods after Book Covers, which removes the new covers. Unlimited Bookshelves changes the objects bounds, its author recommends to load unlimitedbookshelves.esp directly after USSEP and let other mods overwrite it.

masterlist.yaml
```diff
@@ -1346,3 +1346,36 @@ plugins:
...
+  - name: 'Book Covers Skyrim - Lost Library.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/902/'
+        name: 'Book Covers Skyrim on Nexus Mods'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3203152'
+        name: 'BCS Lost Library - Original on Bethesda.net'
+      - link: 'https://bethesda.net/en/mods/skyrim/mod-detail/3203032'
+        name: 'BCS Lost Library - Desaturated on Bethesda.net'
+    tag:
+      - Delev
```
MacSplody, line 1377: Lost Library inserts books into Merchant and Item Levelled Lists.

Would it be possible to detect when a bash patch is required using tags? Or would it be just a guess?
I mean could loot cache crc for bash patch & check if a mod with Delev.Relev,Invent or Names was installed since?
Would probably be out of scope & add to sort running time, which would not be worth it.

Ortham: > Would it be possible to detect when a bash patch is required using tags? Or would it be just a guess?

It would be a guess, and probably not a very reliable one. As for detecting if a Bashed Patch needs a rebuild, that's pretty tricky, especially for pre-Skyrim games where plugin timestamps are modified for load order. I don't think it's worth the additional complexity, though I wouldn't say it's totally out of scope. Just that if you're using a Bashed Patch, the best place for such a feature would be in Bash itself.

PR #81, opened 2018-04-12
Even Better Quest Objectives Patch for AS-LAL

Graybark:


MacSplody (commented):
Just my two cents, but this already loads after main plug-in as it's a master, it's also included in main installer, which will detect if Alt start is installed and add it.

Graybark:
True. Sorry about that. Must have been a very old meta edit I did.

MacSplody:
No problem, don't know if you checked but it only failed because of a syntax error.
```
YAMLException: bad indentation of a sequence entry at line 842, column 8:
        url: 
           ^
```

Graybark:
No, I did not check but I still fail to see what is wrong with the syntax.

MacSplody:
- name  missing colon

Graybark:
Oh. Thanks.

PR #82, opened 2018-04-12
Load ICAO After LAL

Graybark:
Failing that will cause a cyclic interaction between "Bells of Skyrim - ICAIO Patch.esp" and "Alternate Start - Live Another Life.esp". Back cycle: Alternate Start - Live Another Life.esp, Alternate Start LAL TKAA Patch.esp, Immersive Citizens - AI Overhaul.esp, Bells of Skyrim - ICAIO Patch.esp as per master list (no user meta data!).

LAL edits seem to also have been integrated with ICAO since 0.39a: https://www.nexusmods.com/skyrimspecialedition/mods/173?tab=posts

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1221,6 +1221,7 @@ plugins:
         name: 'Immersive Citizens - AI Overhaul [PC] on Bethesda.net'  
     global_priority: 40
     after:
+      - 'Alternate Start - Live Another Life.esp'
       - 'Eli_Breezehome.esp'
       - 'Eli_Routa.esp'
       - 'Eli_Routa - Non-Stormcloak.esp'


Ortham:
Are you sure this is happening with the `v0.10` branch, or are you using a dev build of LOOT that supports group metadata?

MacSplody:
I have tested those mods with Version 0.12.5 and I'm not receiving an error.
Group version does throw an error though. should work fine with 
```
  - name: 'Alternate Start LAL TKAA Patch.esp'
    group: 'Alternate Start'
```

Graybark:
It is a build that supports groups. Had not considered that this might affect it (Version 0.12.5 build 467942d2). I see now that there is a version 0.13 for the new groups version that I should rather have taken a look at.

I still think that ICAIO should load after LAL since the latter does not make any considerations regarding the former and ICAIO even integrates LAL edits. Right now, without any user meta data, LOOT will sort LAL after ICAIO. What do you guys think?

Ortham:
I'm going to close this because the issue described doesn't affect the target branch. I don't have any experience with these specific plugins but if ICAIO incorporates LAL edits, it sounds like the former should load after the latter, and I think this is also the cause of loot/loot-api#21.

What happens if `after` metadata is added to ICAIO to load it after LAL, rather than changing any groups?

Graybark:
`after` metadata fixes it for me in the groups build.

MacSplody:
With just ASLAL & ICAIO, setting ICAIO to load after works fine. With patches it also works.

edit There may be some issues with lighting mods, if ICAIO is set to load after ASLAL though.
As I know ELFX - Hardcore.esp which is set to load after ICAIO, causes an issue for Alt starts starting room. ELFX - Enhancer.esp(some slight light flickers) & ELE_SSE.esp can safely load after ALT start, a few acoustic changes are overwritten.

Ortham:
Reopening this so I don't forget about it...

Ortham:
Actually, closing in favour of a new issue.

PR #83, opened 2018-04-12
Remove Implicit USSEP Bash tags & RWT Clean plug-in info

MacSplody:
Removed some implicit USSEP bash tags

Tags removed were result of changes forwarded from Unofficial Skyrim Special Edition Patch, which was not an explicit master.

Realistic water two
- Added sound tag as it modifies audio. Attenuation Values (ANAM)
- clean plug-in info added.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1021,10 +1021,9 @@ plugins:
       - 'CFTO.esp'
       - 'Honeyside.esp'
     tag:
-      - C.Location
-      - C.Regions
       - C.Water
       - Graphics
+      - Sound
   - name: 'SSEMerged.esp'
     global_priority: 80
     tag:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1111,7 +1101,6 @@ plugins:
       - 'PerkusMaximus_CombatPlus.esp'
     tag:
       - C.Regions
```
CptMcSplody: Modifies a few region names which are unrelated to Update or USSEP so this was fine.

masterlist.yaml
```diff
@@ -1111,7 +1101,6 @@ plugins:
...
-      - C.Water
```
CptMcSplody: Again reverts changes.

PR #91, opened 2018-04-14
Enhanced Lighting for ENB (ELE) - Metadata Update

MacSplody:
- Global Priority 55
- Load After Rules
- Incompatible mods
- Bash tags
- Clean plug-in info

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1408,8 +1408,37 @@ plugins:
     url: 
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/1377/'
         name: 'Enhanced Lighting for ENB (ELE) - Special Edition on Nexus Mods'
+    global_priority: 55
     after:
+      - 'Alternate Start - Live Another Life.esp'
+      - 'Ars Metallica.esp'
+      - 'Book Covers Skyrim - Lost Library.esp'
+      - 'Book Covers Skyrim.esp'
+      - 'EmbersHD.esp'
       - 'Immersive Citizens - AI Overhaul.esp'
+      - 'Inigo.esp'
+      - 'Ivarstead.esp'
+      - 'Karthwasten.esp'
+      - 'LightngDuringStorm.esp'
+      - 'MLU.esp'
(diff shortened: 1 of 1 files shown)


MacSplody:
Global Priority could be set to 70+ and you wouldn't need half the rules, but  at 55 (STEP Guide Recommended Priority) it will overwrite most mods that revert its changes. The following was my load order in testing.

[Testing load order.txt](https://github.com/loot/skyrimse/files/1910726/Testing.load.order.txt)

Without Immersive citizens installed it placed ELE just after Relighting Skyrim. Which meant it was overwritten by the other mods listed in load after.

Ortham:
I'm not fussed about priorities since they'll be gone in the next release. Having explicit rules is a good thing though, it avoids guesswork and reduces the chance of unexpected results.

MacSplody:
No worries, I'll been testing the new version as well, seem to go a lot smoother with less groups for some reason.

Ortham:
What do you mean by "go a lot smoother"? If there's improvements to be made, now's the best time.

MacSplody:
I get less errors when using load after rules with less groups set up.
I just found that if I wanted to load Mod A after mod B. Then mod A would have to be in the group preceding mod B's group. Otherwise If there was a mod C in a group between which was set to load after mod A, then an sorting error would occur. With more groups the chances of an error happening were greater.

This is something the user could fix, but feedback messages could be improved on the errors.

I basically have my groups set as DLC -> Masters -> Low Priorty - > Default -> high priorty -> lateloaders. And treat each as a mini load order.

Ortham:
> With more groups the chances of an error happening were greater.

OK, that's just part of the system enforcing consistency. The alternative (what priorities do) leads to unintuitive and quite complex hierarchies that are relatively tricky to debug.

> This is something the user could fix, but feedback messages could be improved on the errors.

Any suggestions? I find UX/UI things hard, perhaps because as a developer I'll put up with all sorts of arcane and bizarre interfaces.

MacSplody:
I agree with the move to groups, it's a way better system. 

I'll have to do some more testing and see if I can think of anything.

PR #92, opened 2018-04-14
Realistic Water Two - Load After Rules & Global Priority Update

MacSplody:


MacSplody:
I would prefer to set its global priority to 72-75, it does load after Alt start anyway which is 70, and no major issues have been reported. It would avoid most of these load after rules (mainly small compatibility patches, which loot sorts after RWT, such as the farmhouse chimney patches) and it should generally have priority because if it is overwritten the player will have water seams & LOD issues.

Ortham:
I'm not fussed about priorities since they'll be gone in the next release. Having explicit rules is a good thing though, it avoids guesswork and reduces the chance of unexpected results.

PR #101, opened 2018-04-16
Merged v0.10 updates into v0.13

MacSplody:


Ortham:
Yikes, I knew it would be a lot of changes, but didn't realise it would be quite that many. I'll look through it properly later, but did you encounter any merge conflicts?

MacSplody:
16 conflicts fixed, and 4 global priorities which had to be changed to use group name.
No major conflicts, just swapping global for groups really.

PR #102, opened 2018-04-16
V0.13 New Entry - No snow under the roof

MacSplody:
- Dirty Plug-in


Ortham:
I wouldn't make separate commits for the same thing on two different branches, it's better to just merge from one to the other, so closing this in favour of the `v0.10` branch PR.

PR #105, opened 2018-04-18
Warn if SKSE is missing when a SKSE plug-in is detected.

MacSplody:


MacSplody:
Will only show one warning message even if there are multiple .dll. Specific dll checks work fine, but I think there are too many to keep track of & a catch all would be better.

PR #112, opened 2018-04-21
Unofficial Patch - Version Update 4.1.3

MacSplody:
- Clean plug-in crc updated.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -426,7 +426,11 @@ plugins:
         name: 'Unofficial Skyrim Special Edition Patch on Bethesda.net'
     global_priority: -120
     clean:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -426,8 +426,12 @@ plugins:
         name: 'Unofficial Skyrim Special Edition Patch on Bethesda.net'
     global_priority: -120
     clean:
+      # Version 4.1.2a
```
MacSplody, line 429: I meant to ask earlier, is it fine to leave old versions or should these be removed?

Ortham: Old versions should stay, as they might still be available / used.

PR #117, opened 2018-04-25
Load Alt Start after mlu to prevent Cyclic interaction

MacSplody:
Failed to sort plugins.
Details: Cyclic interaction detected between
"imp_helm_legend.esp" and "Alternate Start - Live Another Life.esp".
Back cycle:
Alternate Start - Live Another Life.esp,
BetterQuestObjectives-AlternateStartPatch.esp,
Immersive Citizens - AI Overhaul.esp, CFTO.esp,
MLU.esp,
imp_helm_legend.esp

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -914,6 +914,7 @@ plugins:
     after:
       - 'Open Cities Skyrim.esp'
       - 'My Home Is Your Home.esp'
+      - 'mlu.esp'
     tag:
       - C.Location
   - name: '3rdEraWeapMoS-lvlList.esp'


MacSplody:
I had the same result with a different group order.

https://github.com/CptMcSplody/skyrimse/tree/backup

Edit: 
- Another solution I was testing, was to make sure a mod is only one group before or is in the same group as the mod it's loading after. 
- It appears to have better prevention of further cyclic interaction on addition of new mods to a load order.
https://github.com/CptMcSplody/skyrimse/tree/conflict-res

Ortham:
Does it also work if you load `imp_helm_legend.esp` in the `mluGroup` group instead? I'm not sure if changing MLU's load order is a minimally-changing fix, which would be ideal.

Also, what's causing `imp_helm_legend.esp` to want to load before LAL, is it just their groups?

MacSplody:
no, still get cyclic after putting imp into mlu group. If I move mlu to default group it sorts.

Alternate Start - Live Another Life.esp, (Alt start group)
BetterQuestObjectives-AlternateStartPatch.esp (Default) load after master

Immersive Citizens - AI Overhaul.esp (Immersive Citizens Group) load after Alt start.
CFTO.esp  (Default) load after IC AIO

MLU.esp, (mluGroup) no load after rules
imp_helm_legend.esp (Default) load after mlu.esp

edit: 
Okay, I think your right that its just the groups causing it,
mlu.esp & imp_helm_legend.esp are trying to sort after alt start( or rather Alt & all its load afters), and alt start is trying to load after imp_helm_legend.esp by groups, as there is no load after rules between them, that's why I think having Alt start set to load after mlu.esp fixes the issue, imp helm then seems to share that rule from mlu.esp.

So in testing, if I remember correctly, setting 
BetterQuestObjectives-AlternateStartPatch.esp & CFTO.esp group to a higher priority than imp_helm_legend.esp group also resolves the issue. I'm just going to test to confirm that one.

MacSplody:
Okay tried `imp_helm_legend.esp` in `mlugroup`. Seems okay, another cyclic though. Next is for another mod loading after `mlu.esp`

`"NotSoFast-MainQuest.esp" and "Alternate Start - Live Another Life.esp". Back cycle: Alternate Start - Live Another Life.esp, BetterQuestObjectives-AlternateStartPatch.esp, Immersive Citizens - AI Overhaul.esp, CFTO.esp, MLU.esp, NotSoFast-MainQuest.esp`

move `NotSoFast-MainQuest.esp` to mlugroup. Sorting...

```
 74 4a     LightngDuringStorm.esp
 75 4b     Vivid Weathers SE - Extended Snow.esp
 76 4c     WondersofWeather.esp
 77 4d     Open Cities Skyrim.esp
 78 4e     Alternate Start - Live Another Life.esp
 79 4f     BetterQuestObjectives-AlternateStartPatch.esp
 80 50     FarmhouseChimneysLAL+CRF.esp
 81 51     Immersive Citizens - AI Overhaul.esp
 82 52     Relationship Dialogue Overhaul.esp
 83 53     TheChoiceIsYours.esp
 84 54     Serana Dialogue Edit.esp
 85 55     CFTO.esp
 86 56     Convenient Horses.esp
 87 57     MLU.esp
 88 58     imp_helm_legend.esp
 89 59     NotSoFast-MainQuest.esp
 90 5a     ELE_SSE.esp
 91 5b     RealisticWaterTwo.esp
```

Ortham:
I think a rule of thumb is emerging that's if a plugin A has plugin B in its `after` metadata or as a master, A should be in the same group (or a later-loading group) as B. So `imp_help_legend.esp`  is `after` `MLU.esp`, so should be in `mluGroup` or a later-loading group, and same for `NotSoFast-MainQuest.esp`.

MacSplody:
Would have to have some automatic way to assign a group by master or load after rule. And also prevent the user moving it below this assignment( without first changing the parent to a lower group), but also allow them to move either to higher group.

Ortham:
Yes, but that sort of inheritance introduces long-distance complexity as mentioned in loot/loot-api#22, and is one of the reasons priorities are so difficult to reason about.

MacSplody:
I'd like to change the groups, as the way they are set up will also cause issues.

Currently there are two `floating` groups in `MLUgroup & immersivecitizensgroup`. 
Now by `floating`, I mean there is nothing to stop mods in those groups loading last unless there is a mod that must load after them. In my example above `ELE.esp` & `realisticwatertwo.esp` are both set to load after `mlu.esp`. `mlu.esp` reverts changes made by those mods if loaded after them.

> if a plugin A has plugin B in its after metadata or as a master, A should be in the same group (or a later-loading group) as B

I think with this rule, there should be no need for the two floating groups.

Another thing I noticed in testing, when two mods are in the same group, it can make some load after rules redundant, as loot will generally sort two mods correctly when there are less external influences on comparing those two mods, which groups reduce.

Ortham:
`mluGroup` exists because `MLU.esp` must load after other plugins unless otherwise specified. There aren't any groups that load after `mluGroup` because no such relation has been defined, so it's 'floating' as you say, but I don't see how implementing group inheritance will negate the need for `mluGroup` to ensure `MLU.esp` loads after other plugins in general.

I'll have a go at implementing group inheritance when I get the time, then we can evaluate the pros/cons against real load orders.

MacSplody:
It doesn't negate the need for `mluGroup` ,what I meant is with the above rule you mentioned, it negates the need to have it, for lack of a better phrase 'floating'. It can be above default and not cause any issues if every mod after it, is of the same or higher group.

Ortham:
Ok, I understand now

PR #124, opened 2018-05-14
Added MorrowLoot Ultimate Patches

Graybark:
To prevent cyclical interaction.


Graybark:
Tested it internally. No effect.

PR #126, opened 2018-05-27
Add tags for Diverse Skyrim SSE

spaceoden:
adds Diverse Skyrim SSE and the relev and delev tags needed for wrye bash to respect its list changes. No sorting data because the mod has no known conflicts.


Ortham:
Thanks! 👍

Issue #130, opened 2018-06-03
LOOT 13.1 Cyclic Interaction


missnasty07:
[LOOTDebugLog.txt](https://github.com/loot/skyrimse/files/2066211/LOOTDebugLog.txt)

It says a cyclic interaction detected between RealisticWaterTwo++Verdant+Patch.esp and ELE_SSE.esp. Back cycle: ELE_SSE.esp, Qw_ELE_CRF Patch.esp, RealisticWaterTwo.esp, RealisticWaterTwo++Verdant+Patch.esp.

Ortham:
Try using the metadata editor to put `Qw_ELE_CRF Patch.esp` in the `Immersive Citizens` group or `RealisticWaterTwo+-+Verdant+Patch.esp` in the `Alternate Start` group.

missnasty07:
Okay the Qw_ELE_CRF Patch.esp change to immersive citizens group does not work, still gives error, but the RealisticWaterTwo++Verdant+Patch.esp changed to Alternate Start group worked.
Thank you for the help!

PR #131, opened 2018-06-03
Merge immersive Citizens group into Alt Start Group.

MacSplody:
- ELE_SSE & ICAIO are the only mods in immersiveCitizensGroup, both are set to load after alt start.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -226,9 +226,6 @@ groups:
   - name: &raceGroup Races
     after: [ default ]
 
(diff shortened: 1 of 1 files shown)


MacSplody:
It should resolve these issues.
#130 
https://github.com/loot/loot/issues/968
https://github.com/loot/loot/issues/963

Ortham:
> ELE_SSE & ICAIO are the only mods in immersiveCitizensGroup, both are set to load after alt start.

Well, I can't fault the logic of this then...

Issue #141, opened 2018-06-06
Yet Another Cyclic Problem


gabe50200:
Seems to be same problem, but different .esp's:
Cyclic interaction detected between "JKs Skyrim_RWT_Patch.esp" and "ELE_SSE.esp". Back cycle: ELE_SSE.esp, Immersive Citizens - ELE patch.esp, RealisticWaterTwo.esp, JKs Skyrim_RWT_Patch.esp
[LOOTDebugLog.txt](https://github.com/loot/skyrimse/files/2074433/LOOTDebugLog.txt)
Apparently the Immersive Citizens fix didn't work for me, although it still seems to be involved.

MacSplody:
Hi @gabe50200 can you change `Immersive Citizens - ELE patch.esp` and `JKs Skyrim_RWT_Patch.esp` group to to Alternate Start?

![image](https://user-images.githubusercontent.com/1944639/41010466-8ff5f3a4-692f-11e8-8cd1-79c44a90b01b.png)

MacSplody:
@gabe50200 Is this issue still occurring?

gabe50200:
Bingo, that did it. Thanks much.

Ortham:
Closing this then.

PR #142, opened 2018-06-06
[Meta Data] Move RWT to higher group than Alt Start

MacSplody:
Add New Group - Late Loaders after alt start.

Move RealisticWaterTwo.esp to new group.

`Failed to sort plugins. Details: Cyclic interaction detected between "JKs Skyrim_RWT_Patch.esp" and "ELE_SSE.esp". Back cycle: ELE_SSE.esp, Immersive Citizens - ELE patch.esp, RealisticWaterTwo.esp, JKs Skyrim_RWT_Patch.esp`

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -240,8 +240,11 @@ groups:
   - name: &alternateStartGroup Alternate Start
     after: [ *mluGroup ]
 
-  - name: &dynamicPatchGroup Dynamic Patches
+  - name: &lateLoadersGroup Late Loaders
     after: [ *alternateStartGroup ]
+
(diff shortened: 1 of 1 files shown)


MacSplody:
There have been repeated cyclic errors between `ELE_SSE.esp` & 'RealisticWaterTwo.esp' related patches.
A load after rule is not required as the conflicts only cause minor visual errors which require a small patch to resolve.

RealisticWaterTwo losing the conflict, causes a LOD to not generate correctly. It can also cause water seams which are more noticeable.
Best option is to place 'RealisticWaterTwo.esp' in a group above `ELE_SSE.esp'.

#141

PR #156, opened 2018-06-09
Merge v0.13 changes to v0.10

MacSplody:


Ortham:
This breaks priorities for v0.10 users as it replaces all of them with v0.13 groups.

EDIT: Or it would, if it was merged into the v0.10 masterlist branch, I didn't spot it wasn't the target branch.

MacSplody:
Just being extra careful.

PR #157, opened 2018-06-09
Merge v0.13 updates to v0.10

MacSplody:


MacSplody:
I'd prefer someone give it a review since its a large commit. If your happy enough with that it can be merged.

masterlist.yaml
```diff
@@ -198,7 +203,7 @@ globals:
 # SKSE64 plugins requirement
   - <<: *requireSKSE64
     subs: [ 'a SKSE Plugin' ]
-    condition: 'file("SKSE/Plugins/([^\.]+\.dll)") and not file("../skse64_loader.exe")'
+    condition: 'file("SKSE/Plugins/([^\.]+\.dll)") and not file("../skse64_loader.exe") and not file("../sksevr_loader.exe")'
```
MacSplody, line 206: Are the VR checks even required?

masterlist.yaml
```diff
@@ -324,6 +356,14 @@ plugins:
         itm: 69
         udr: 8
         nav: 1
+    clean:
+      # Version 1.59.39.0 - Auto Clean
+      - crc: 0x5eb9eade
+        util: SSEEdit v3.2.1
+  - name: 'SkyrimVR.esm'
+    global_priority: -127
```
MacSplody, line 364: VR again, remove?

masterlist.yaml
Ortham: Yes, if it's not there then SKSEVR users will see messages about not having SKSE installed.

masterlist.yaml
Ortham: Nope, Skyrim VR uses the same masterlist because there's so little difference, and `SkyrimVR.esm` needs to load early.

Ortham (approved):
Looks good to me.

MacSplody:
Okay, thanks.

PR #165, opened 2018-06-14
Add check for merged patch to MLU patch conditionals

MacSplody:


masterlist.yaml
```diff
@@ -1754,7 +1755,6 @@ plugins:
       - 'PerkusMaximus_Warrior.esp'
       - 'PerkusMaximus_CombatPlus.esp'
     tag:
-      - C.Regions
```
Ortham, line 1757: Was removing this intentional?

MacSplody: Yes, it's a bad tag caused by a forwarded change from the unofficial patch. I forgot to add a note about it.

Ortham: OK

PR #168, opened 2018-06-15
New entry - Whiterun Hold Forest, Update Touring Carriages

MacSplody:
### Whiterun Hold Forest
- Set to early group as it should be overwritten.
- Added cleaning info

### TouringCarriages
- Added cleaning info


MacSplody:
Any ideas as to why this failed? I've committed again on a new branch and it passed, strange.

Ortham:
Looks like an issue with Travis.

PR #181, opened 2018-06-18
[New Groups] Core Mod Groups

MacSplody:
- Add Base Node for Core mod Groups
-  mod groups for Alt Start, Open Cities & Immersive citizens
image

Groups load after base node, so they have greater priority.
Each group added must load before and after at least one group member to avoid conflicts.
Chain must end in another group node to maintain structure & priority.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -239,8 +239,20 @@ groups:
   - name: &lowPriorityGroup Low Priority Overrides
     after: [ default ]
 
-  - name: &highPriorityGroup High Priority Overrides
+  - name: &coreGroup Core Mods
     after: [ *lowPriorityGroup ]
+
+  - name: &ocsGroup Open Cities
+    after: [ *coreGroup ]
+
+  - name: &altStartGroup Alternate Start
+    after: [ *coreGroup,*ocsGroup ]
+
+  - name: &iCitizenGroup Immersive Citizens
+    after: [ *coreGroup, *ocsGroup ]
+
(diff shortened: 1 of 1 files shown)


Ortham:
There's no point to the chain you've created, the RDO group will load after High Priority Overrides even without the direct load after edge because there's a path going through IC, AS and OC.

MacSplody:
Okay I just thought it would be easier for the user to follow.
image


I'll remove them.

Ortham:
Well, I think having more lines is more confusing... A single path through the group nodes looks simpler to me.

MacSplody:
It was mainly a way to handle patches and load after rules, I wanted it to be possible to give each mod its own group as quickly as possible.
My idea was to use groups such as Core Mods & Low Priority as primary nodes. 
They would simply act as the priority anchors for all groups the user could add, and since users added groups looked like this I felt it should be consistent.
But if it sounds too messy, or could could cause any issue then I'll take it out.

And it also made it easier to add groups without breaking the existing groups.

Ortham:
I'm fine with having Core Mods and Low Priority groups, but the messy bit of the graph above is functionally identical to:

Core Mods -> Open Cities -> Alternate Start -> Immersive Citizens -> High Priority Overrides -> Late Loaders -> Enhanced Lighting -> Realistic Water -> Dynamic Patches

Adding in extra lines everywhere gains you nothing but visual and computational complexity.

MacSplody:
#187 Okay if it's making sorting less efficient and also harder to understand then I'll use a straight line.

PR #185, opened 2018-06-20
[Format] Reorganize Masterlist

MacSplody:
Mods are now listed by a general category & A-Z. 
Main reason is to make it easier for anyone contributing.
- They should now be able to quickly add an entry were a similar category mod or plug-in from the same mod exists.


MacSplody:
Changed my mind, I'll do any formatting in smaller chunks, easier to review and safer.

PR #194, opened 2018-06-21
Remove Section Headers

MacSplody:


Ortham:
OK, so to expand upon my previous comment:

It's nice when similar plugins are co-located in the masterlist, but reorganising them into themed categories has the following problems:

1. It makes it more difficult to track changes. You can use `git blame -L <start line>,<end line> masterlist.yaml` to see what commit last changed each line, and who made the commit, and this is great for seeing when metadata was added, etc. If you move entries around though, git blame will only show you when the entry was moved, then you've got to check that commit to see the old location, and do another git blame from that commit's parent over the old location line range.
2. A plugin may seem like it belongs in more than one category, so whichever one you pick, it's going to confuse someone else who thinks the another category would be a better choice. This can lead to plugins being moved back and forth by different people, perhaps totally unknowingly.
3. It can make it seem like the order of entries in the masterlist matters, when it doesn't, potentially tripping up new contributors.

Given that moving plugins around doesn't actually *do* anything, I don't think it's worth those issues. That said, you've gone to the effort of manually moving a lot of entries around, so I'm interested in your motivation to see if there's an alternative solution.

MacSplody:
Well the main reasons were for clarity and maintenance.
I agree that contributors could get tripped up, but I also think having some form of organization could also help them.
I had hoped to keep the themed categories as generic as possible, as I agree that it can be quite subjective to order mods that way if they are too specific. I was using them more as guideline to keep similar mods together.
This ordering does work better in some cases such as audio, weather & lighting, but not so much for game-play ones. I could have split the game-play section into several different areas as those are so varied in scope.
Having a section for some mods where its more clear cut, such as the script/utility generated files, shouldn't really cause any issues for contributors though.
But I do agree with your points, if it causes too many issues to maintain, then it's not worth it for saving some time.

MacSplody:
Closing as I want to delete the branch.

Issue #195, opened 2018-06-22
Weather mods


MacSplody:
~~## [Vivid Weathers Definitive Edition](https://www.nexusmods.com/skyrimspecialedition/mods/2187/)~~ Updates  ADDED #196 

### Compatibility info
Vivid Weathers have built-in compatibility with:
 - Frostfall

Vivid Weathers is compatible with Interior Lighting mods such as:
 - Enhanced Lighting for ENB - ELE
 - Realistic Lighting Overhaul (do NOT use the weather module with VW)
 - ELFX (do NOT use the weather module with VW)

The following mods offer compatibility patches for Vivid Weathers on their own page:
 - Darker Nights

Vivid Weathers is not compatible with other weathers overhaul

Make sure you load Vivid Weathers after any mod altering Ambient Sounds and Weather Distribution. 
~~Load Vivid Weathers after Wet & Cold (for weather distribution in Solstheim).~~

###### Load After
~~[Wet & Cold](https://www.nexusmods.com/skyrimspecialedition/mods/644)~~
Patch required.
###### Load Before
[A Quality World Map](https://www.nexusmods.com/skyrimspecialedition/mods/5804)

##### Incompatible mods

- [Aequinoctium.esp](https://www.nexusmods.com/skyrimspecialedition/mods/15884)
- [Autumn Skies.esp](https://www.nexusmods.com/skyrimspecialedition/mods/14379)
- [ClimatesOfTamriel.esm](https://www.nexusmods.com/skyrimspecialedition/mods/2237)
- [Dolomite Weathers.esp](https://www.nexusmods.com/skyrimspecialedition/mods/7895)
- [ELFX - Weathers.esp](https://www.nexusmods.com/skyrimspecialedition/mods/2424)
- [Foggy Weather 2.esp](https://www.nexusmods.com/skyrimspecialedition/mods/6421)
- [Mythical Ages.esp](https://www.nexusmods.com/skyrimspecialedition/mods/11578)
- [NAT.esp](https://www.nexusmods.com/skyrimspecialedition/mods/12842)
- [NorthernSWS.esp](https://www.nexusmods.com/skyrimspecialedition/mods/14878)
- [Obscura Weather and Lighting.*\.esp](https://www.nexusmods.com/skyrimspecialedition/mods/12523)
- [Obsidian Weathers.esp](https://www.nexusmods.com/skyrimspecialedition/mods/12125)
[...]

PR #196, opened 2018-06-22
Weather Mods & True Storms Compatibility Patches

MacSplody:
#195

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1181,11 +1181,36 @@ plugins:
         condition: 'active("TrueStormsSE.esp") and not active("Obsidian_TS_Patch_.*\.esp") and not active("True Storms - Obsidian Weathers - Patch .esp")'
         display: 'True Storms Special Edition'
       - 'Vivid WeathersSE.esp'
+    msg:
+      - <<: *hasPluginPatch
+        subs:
+          - '[True Storms SE](https://www.nexusmods.com/skyrimspecialedition/mods/2472/)'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -32,7 +32,7 @@ common:
     type: warn
     content:
       - lang: en
-        text: "You seem to be using %1%, but you haven't enabled the %1% patch for this plugin: %2%"
+        text: "You seem to be using %1%, but you haven't enabled the patch for this plugin: %2%"
```
MacSplody, line 35: Not sure about this, but I think its easier to read.

> You seem to be using True Storms SE, but you haven't enabled the patch for this plugin: NAT - TrueStorms Merged Compatibility SSE

> You seem to be using True Storms SE, but you haven't enabled the True Storms SE patch for this plugin: NAT - TrueStorms Merged Compatibility SSE

'You seem to be using' Doesn't sound very certain. 
Maybe something like 

> This mod conflicts with [Mod X](url). LOOT has not detected a known compatibility patch. A patch can be found here: [compatibility patch](url). Please report dead links and any corrections to the [LOOT team](url).

Ortham: I think the new wording is better, and I like "You seem to be using" just in case they're not actually using that mod, but another one with the same plugin name (which happens). Better to sound a little unsure than to say "You are using X" and have the user respond with "uh, no I'm not".

Ortham (approved):

PR #202, opened 2018-06-30
add TheWulfPanda's teleport spell mods to masterlist

jb29:
Hi! Here's the second mod of the Teleport Spells: [Teleport - Faction Spells](https://www.nexusmods.com/skyrimspecialedition/mods/6991/)

I guess you could argue about the cleaning instructions because the last item in the list isn't a wild edit but a fix from [USSEP](https://www.nexusmods.com/skyrimspecialedition/mods/266). I thought there's no harm in adding it anyway.

The author added the flag `Don't Havok Settle` on the following references:
- `[REFR:02010521]`
- `[REFR:000CA72C]`
- `[REFR:000B3434]`

I don't understand why mists and walls would benefit from this flag, so I removed those references too.

**Edit:**
I added the last teleport mod, too: [Teleport - Towns and Villages](https://www.nexusmods.com/skyrimspecialedition/mods/7267/)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3046,6 +3046,64 @@ plugins:
         util: SSEEdit v3.2.1
       - crc: 0x7BD6C78A # v1.03 Without wild edits
         util: SSEEdit v3.2.1
+  - name: 'TheWulfPanda - Teleport Spells Towns and Villages.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/7267/'
+        name: 'Teleport - Towns and Villages on Nexus Mods'
+    inc:
+      - 'TheWulfPanda - Teleport Spells Towns and Villages Less Magicka Increase Cast Time.esp'
+    msg:
+      - <<: *hasWildEdits
+        subs:
+          - '[TheWulfPanda - Teleport Spells Towns and Villages.esp](https://www.nexusmods.com/skyrimspecialedition/mods/7267/)'
+          - |+
+
+            #### Remove the following records with xEdit
+            - `DragonBridgeExterior03 [CELL:00009349]`
+            - `Riverwood [CELL:00009732]`
+            - Parent Sub-Block 2,-3 of `IvarsteadExterior02 [CELL:000097BF]`
+            - Remove the **XCWT** attribute of `Riverwood04 [CELL:00009731]`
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1975,6 +1975,9 @@ plugins:
         util: SSEEdit v3.2.1
   - name: 'Ars Metallica.esp'
     url: ['https://www.nexusmods.com/skyrimspecialedition/mods/321/']
+    after:
+      - 'TheWulfPanda - Faction Teleport Spells.esp'
+      - 'TheWulfPanda - Faction Teleport Spells Less Magicka Increased Cast Time.esp'
```
MacSplody: These load after rules are not needed as its only a weather region edit that is being overwritten, which isn't critical to `Ars Metallica.esp`

image

MacSplody (changes requested):
Everything else is fine, just remove the load after rules for Ars Metallica.

masterlist.yaml
jb29: Okay, I rebased my PR.

MacSplody (approved):

jb29 (commented):
I'm not sure if assigning me means I should leave a review. So here it is. 😄 

The text is actually indented with 4 spaces now while everything else uses an indentation of 2 spaces.
But in case of multiline strings this is just cosmetical anyway. YAML ignores the indentation in multiline strings, so you could even write them without an indentation at all as long as the next key has the correct indentation again.
So technically, both versions are fine.

MacSplody:
hmm... Are you not able to merge this? I must be remembering incorrectly, I could have sworn that I was able to merge after being approved.

jb29:
No, I don't have the rights to do that.
I'm not even allowed to approve or request changes in a review, I can just comment them. 😄

MacSplody:
Github is strange, why have approval if a team member has to merge it anyway.

Ortham:
> Github is strange, why have approval if a team member has to merge it anyway.

It's probably because you can configure a repository to require reviews before a PR can be merged, though none of LOOT's repos are like that.

jb29:
Some projects require multiple reviews, for example from a frontend and backend developer.

And in some commercial projects the PRs will be rolled out to a staging system after approval where they will be further tested by their stakeholders before they're finally allowed to be merged to the master.

PR #203, opened 2018-06-30
update 'Better Vampires' list entry

jb29:
Version 8.2 of [Better Vampires](https://www.nexusmods.com/skyrimspecialedition/mods/1925) is clean. The Bash Tags were generated with [Generate Tags for Wrye Bash.pas](https://github.com/fireundubh/xedit-scripts/blob/master/all/Generate%20Tags%20for%20Wrye%20Bash.pas).

There's a minor conflict with [Tainted Blood](https://www.nexusmods.com/skyrimspecialedition/mods/9312).
Tainted Blood removes a race check from the Night Cloak spell while Better Vampires completely overhauls this spell. So Better Vampires should have precedence over Tainted Blood and should be loader after it.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1991,6 +1991,16 @@ plugins:
     after:
       - 'EBM_UnlimitedPower.esp'
       - 'EBM_UnlimitedPower_Vampire.esp'
+      - 'Tainted_Blood.esp'
+    tag:
+      - Graphics
+      - Names
+      - Relev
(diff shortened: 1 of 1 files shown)


MacSplody (commented):
Yeah that looks good, I'm going to add a commit to this involving Sacrosanct compatibility. I've been meaning to add it for a while.

MacSplody (approved):

PR #206, opened 2018-07-01
add 'A Quality World Map' to masterlist

jb29:
The various versions of [A Quality World Map](https://www.nexusmods.com/skyrimspecialedition/mods/5804/) are all clean but the ESPs are copied over from Oldrim. I wanted to add the CRC for the converted plugins too but then I realized that saving them in the CK changes the file's modification date so the CRC will be different for everyone.

I think™ it should be safe to load the Oldrim versions in SE because the plugins don't touch NPC data or stuff like that.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -812,6 +812,36 @@ plugins:
 #############################
 ## AudioVisual - Interface ##
 #############################
+  - name: 'IcePenguinWorldMap.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/5804/'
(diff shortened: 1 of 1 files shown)


MacSplody:
Last time I checked, it was only the latest vivid versions that were form 43. The rest are 44, including the paper one. I don't think there is any issue with adding the CRC for the vivid versions, even if they are form 43.

Edit Typed in a hurry, it may be slightly easier to read now.

jb29:
Oh, my bad. Yes, your're right.
Only the latest version is outdated. (That's a strange sentence 😄)

MacSplody (approved):

Issue #217, opened 2018-07-08
Legacy of the Dragonborn has an Open Cities Skyrim compatibility patch


ghost:
Legacy of the Dragonborn : https://www.nexusmods.com/skyrimspecialedition/mods/11802
It is under the optional Legacy of the Dragonborn Patches SE file.

Open Cities Skyrim : https://www.nexusmods.com/skyrimspecialedition/mods/281/

The patch ESP is : DBM_OpenCities_Patch.esp

bpipher0225:
LOTD is probably worth a group on its own, let me know what info you might need and I could even write up something to handle most of the legwork.

PR #218, opened 2018-07-08
New Entry - Legacy of the Dragonborn

MacSplody:
#217 @bpipher0225 wants to contribute, so I thought I'd get it started for them.

Add suggested bash tags
- Graphics, stats & sounds as it modifies matching records for a lot of items.

Gave it a separate section as it's quite a large mod & has over 70+ patches and a few add-ons.
I don't plan on adding all these patches, just ones that require it.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -4025,7 +4025,6 @@ plugins:
         name: 'Legacy of the Dragonborn Patches SE on Nexus Mods'
     tag:
       - Actors.AIPackages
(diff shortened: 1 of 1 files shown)


MacSplody:
@bpipher0225 Thanks for the contribution. Don't get to worried about any mistakes, just make a habit of double checking before committing. Also if your adding any incompatibility or load after rules it helps to have a source or something for future reference.

PR #219, opened 2018-07-08
Legacy of the dragonborn - Patch updates

ghost:
- Added Available Patch messages
- Included Open Cities Skyrim patch
- Added Delev bash tag to DBM_CRF_Patch.esp as according to SSEEdit
- Atom deleted a few hanging spaces

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3986,7 +3986,7 @@ plugins:
   - name: 'LegacyoftheDragonborn.esm'
     url: ['https://www.nexusmods.com/skyrimspecialedition/mods/11802/']
     msg:
-  - <<: *hasPatchIncluded
+      - <<: *hasPatchIncluded
         subs:
           - '[Cutting Room Floor](https://www.nexusmods.com/skyrimspecialedition/mods/276/)'
           - '[Legacy of the Dragonborn SSE](https://www.nexusmods.com/skyrimspecialedition/mods/11802)'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -345,7 +345,7 @@ plugins:
         itm: 621
         udr: 82
         nav: 57
-      - <<: *dirtyPlugin
+      - <<: *dirtyPlugin 
         crc: 0x6f162d5e # v1.59.39.0 - Auto Clean First pass
         itm: 6
         udr: 0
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -3985,6 +3985,27 @@ plugins:
...
+      - <<: *hasPatchIncluded
+        subs:
+          - '[Skyrim Underground](https://www.nexusmods.com/skyrimspecialedition/mods/131/)'
+          - '[Legacy of the Dragonborn SSE](https://www.nexusmods.com/skyrimspecialedition/mods/11802)'
+        condition: 'active("AKSkyrimUnderground.esp") and not active("DBM_SkyrimUnderground_Patch.esp")'
+      - <<: *hasPatchIncluded
+        subs:
+          - '[Open Cities Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/281/)'
+          - '[Legacy of the Dragonborn SSE](https://www.nexusmods.com/skyrimspecialedition/mods/11802)'
+        condition: 'active("Open Cities Skyrim.esp") and not active("DBM_OpenCities_Patch.esp")'
```
MacSplody, line 4008: Those look fine

masterlist.yaml
```diff
@@ -4035,6 +4057,10 @@ plugins:
     clean:
       - crc: 0xAD717340 # v4.1
         util: SSEEdit v3.2.2
+  - name: 'DBM_OCS_Patch.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/11802/'
+        name: 'Legacy of the Dragonborn Patches SE on Nexus Mods'
```
MacSplody, line 4063: Generally there should be some other information besides name and URL to warrant adding an entry.

ghost: I just checked SSEEdit and no tags were recommended so that can be deleted then.  I'll keep that in mind moving forward, thank you for the feedback!

masterlist.yaml
```diff
@@ -4004,6 +4025,7 @@ plugins:
         name: 'Legacy of the Dragonborn Patches SE on Nexus Mods'
     tag:
       - Actors.AIPackages
+      - Delev
```
MacSplody, line 4028: This delev tag isn't needed, the lvl list record in this patch merges the level list changes from LOTD & Cutting Room floor. 
Delev tag is used when an entry is removed. 
Just remember at xEdit startup you should only tick the plug-in you want to check for bash tags.
As checking with other plug-ins can give you an incorrect tag suggestion.

masterlist.yaml
MacSplody: I'll be adding to it, so it can stay.

PR #222, opened 2018-07-10
Unofficial High Definition Audio Project

MacSplody:
[Unofficial High Definition Audio Project](https://www.nexusmods.com/skyrimspecialedition/mods/18115)

Uses empty Light Master esp to load bsa archives.

```
Skyrim.esm
Update.esm
Dawnguard.esm
HearthFires.esm
Dragonborn.esm
UHDAP - MusicHQ.esp OR UHDAP - MusicUHQ.esp
UHDAP - en0.esp // replace en0 as necessary with respective language
UHDAP - enX.esp // replace X with the rest of the language .esp
... // Any other mods that the authors state should be above USSEP
Unofficial Skyrim Special Edition Patch.esp
... // Rest of your load order
```

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1326,9 +1326,10 @@ plugins:
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/18115/'
         name: 'Unofficial High Definition Audio Project on Nexus Mods'
-    inc:
-      - 'UHDAP - MusicUHQ.esp'
     group: *eslmGroup
+    inc:
+      - 'UHDAP - MusicUHQ1.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -221,9 +221,12 @@ globals:
 groups:
   - name: &dlcGroup DLC
 
-  - name: &earlyLoadersGroup Early Loaders
+  - name: &eslmGroup Light Masters # Non DLC - Has Header Flags (ESM & ESL)
```
MacSplody, line 224: I added a separate group as I think it's better to keep these in a separate group to official esl DLC files, as their behaviour is slightly different.

PR #239, opened 2018-07-14
[Merge Updates] V0.13 dev masterlist

MacSplody:


masterlist.yaml
```diff
@@ -2993,6 +3172,33 @@ plugins:
     clean:
       - crc: 0xDD84DD4C # v2.1
         util: SSEEdit v3.2.1
+  - name: 'Cloaks.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6369/'
+        name: 'Cloaks of Skyrim on Nexus Mods'
+    tag:
+      - name: Delev
+        condition: 'active("Unofficial Skyrim Special Edition Patch.esp")'
```
MacSplody: Not sure if this is good practice, but it was an unusual issue where a levelled list entry removed by USSEP was being added back in by Cloaks.
Even though Cloaks has no Delev or Relev tags.
The expected bash output is only achieved if this Delev tag is present.

masterlist.yaml
```diff
@@ -2993,6 +3172,33 @@ plugins:
...
+      - Invent
+    msg:
+      - <<: *hasPatchIncluded
+        subs:
+          - '[Unofficial Skyrim Special Edition Patch](https://www.nexusmods.com/skyrimspecialedition/mods/266/)'
+          - '[Cloaks - USSEP Patch](https://www.nexusmods.com/skyrimspecialedition/mods/11802)'
+        condition: 'active("Unofficial Skyrim Special Edition Patch.esp") and not active("Cloaks - USSEP Patch.esp") and not active("Bashed Patch, 0.esp")'
```
MacSplody, line 3225: For plug-ins that can be merged and will therefore not be active, can a conditional with file be used instead?

`condition: 'active("Unofficial Skyrim Special Edition Patch.esp") and not file("Cloaks - USSEP Patch.esp")'`

masterlist.yaml
```diff
@@ -3021,9 +3246,30 @@ plugins:
...
+      - Names
       - Relev
+      - Stats
+    clean:
+      - crc: 0x20AA5984 # v1.06a
+        util: SSEEdit v3.2.2
+  - name: 'Immersive Weapons.esp'
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/5336/']
+    after:
+      - AmazingFollowerTweaks.esp
```
MacSplody, line 3301: Taken from LE Master list.

masterlist.yaml
```diff
@@ -3077,6 +3332,61 @@ plugins:
...
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/9660/']
+    tag:
+      - Delev
+      - Relev
+    clean:
+      - crc: 0x0D7685E5 # v1.4a
+        util: SSEEdit v3.2.2
+  - name: 'Lore Weapon Expansion - Daedric Crescent.esp'
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/9660/']
+    group: *earlyLoadersGroup
```
MacSplody, line 3382: Main file is not a master of optional files.
They're loaded early to avoid reverting Cell changes.

masterlist.yaml
```diff
@@ -3077,6 +3332,61 @@ plugins:
...
+  - name: 'Summermyst - Enchantments of Skyrim.esp'
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/6285/']
+    inc:
+      - 'Animated Weapon Enchants.esp'
+      - 'Classic Elder Scrolls Attributes.esp'
+      - 'SkyRe_Main.esp'
+      - 'Wintermyst - Enchantments of Skyrim.esp'
+      - 'Worlds Dawn.esp'
+    after:
+      - 'Weapons & Armor Fixes_Remade.esp'
```
MacSplody, line 3407: Not out for SSE yet but it is being worked on, also easy to convert so it's probably being used. 
It's being merged with Clothing & Clutter fixes so the name will probably be different.

masterlist.yaml
```diff
@@ -3077,6 +3332,61 @@ plugins:
...
+      - <<: *hascompatibilityIssues
+        subs:
+          - 'Summermyst - Enchantments of Skyrim.esp'
+          - 'EnchantingAwakened.esp'
+          - 'Semi-Compatible. Both mods can be active at the same time, but the enchantment limits in Enchanting Awakened do not apply to Summermyst enchantments.'
+        condition: 'active("Summermyst - Enchantments of Skyrim.esp") and active("EnchantingAwakened.esp")'
+      - type: say
+        content:
+          - lang: en
+            text: '[Summermyst Compatibility Notes](https://www.nexusmods.com/skyrimspecialedition/articles/116/?)'
```
MacSplody, line 3423: I think it's useful to add a message link If the compatibility notes are on a separate page or site to the mod description.

masterlist.yaml
```diff
@@ -649,6 +649,7 @@ plugins:
     group: *fixesResourcesGroup
     after:
       - 'Lanterns Of Skyrim - All In One - Main.esm'
+      - 'RSkyrimChildren.esm'
```
MacSplody, line 652: Recommended to load `RSkyrimChildren.esm` before USSEP on RS Children mod page description.

masterlist.yaml
Ortham: So when the Bashed Patch is built with USSEP tagged as Delev and Cloaks without Delev, the Bashed Patch wasn't including the USSEP change? That's unexpected...

masterlist.yaml
Ortham: Yes, if the plugin is fully merged into the Bashed Patch, that makes sense.

masterlist.yaml
Ortham: Yep, I can see that being helpful (I forget articles exist because they're relatively new).

Ortham (approved):
Other than my question about the Delev tag for Cloaks.esp (which isn't a blocker), this looks good.

masterlist.yaml
MacSplody: No my mistake it was Skyrim Immersive creature which added it back in, that has Delev & Relev so  it's expected behaviour.

image

Issue #259, opened 2018-07-26
Repositioning for AS-LAL and OCS


Arthmoor:
I need to ask for a repositioning on Alternate Start - Live Another Life and Open Cities Skyrim.

I've issued updates to each one to move OCS below LAL in the load order in order to preserve region and location data in OCS that LAL was blocking and has been blocking for a long time. This was done to resolve the only incompatibility between the 2 mods and involved having to change scripting on a single marker which was in the way for each one.

I'm not entirely sure what the current list structure looks like but LAL should be moved directly underneath any section that contains village mods, but before any city overhauls. That may be a bit of a special space, but it's probably what will work best for LAL.

OCS should remain as far down the list as it can go.

MacSplody:
@Arthmoor Current order is  

City & village Overhauls (Default Group) -> OCS -> Alt Start.

So if I'm following correctly the new order would be as follows.

City Overhauls ->  Alt Start LAL -> Village Overhauls ( JK Riverwood, Dawnstar etc.? ) -> OCS

Arthmoor:
It should become:

Village Overhauls (Mine, JKs, ETaC, etc) -> Alt Start -> City Overhauls (The 5 walled capitals) -> OCS

MacSplody:
@Arthmoor Change has been submitted in #262. Just waiting on some feedback from @WrinklyNinja

PR #262, opened 2018-07-27
V0.13 staging - Merges Updates

MacSplody:
[Update] Load OCS after alt start (#261) 
#259 Requested change.

[New] SAFO & CSOS (#260)

* [New] Skyrim Alchemy and Food Overhaul (#256)

Source for rules:
Mod description page

Load After:
- Better food effects.
Incompatible with:
- Cooking in Skyrim, both do the same thing.
Added bash Tag suggestions.

Added patch check & messages:
- Another Sorting Mod
- AWESOME POTIONS
- Frostfall
- Imperious
- iNeed
- Realistic Needs and Diseases
- SkyTEST - Realistic Animals and Predators
- Creation Club Survival Mode
- Valdacil's Item Sorting

Added cleaning info for Main plug-ins.

Ordinator must load after SAFO, Perk conflict.

* [New] Carriage Stops of Skyrim (#257)

Load Order
Put it high in the load order. If you have any water, weather, outdoor lighting mods, put it above those.
[Update] Load OCS after alt start (#261)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2729,6 +2729,7 @@ plugins:
     after:
       - 'Phenderix Magic Evolved.esp'
       - 'Phenderix Magic World.esp'
+      - 'SAFO.esp'
       - 'Undeath.esp'
     inc:
       - 'ScopedBows_EagleEyeTweak.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -255,18 +255,18 @@ groups:
   - name: &coreGroup Core Mods
     after: [ *lowPriorityGroup ]
 
-  - name: &ocsGroup Open Cities
-    after: [ *coreGroup ]
-
   - name: &altStartGroup Alternate Start
-    after: [ *ocsGroup ]
+    after: [ *coreGroup ]
```
MacSplody, line 259: I'm not sure what effect swapping group order will have on current users load order.
In hindsight it my be better to not have a group named after a mod, so that it isn't intrinsically linked.

Ortham: It's better to be conservative about making changes because of this, but at least the group names haven't changed, so load orders will change but no user metadata should break (apart from possibly introducing new cycles).

Ortham (approved):

PR #285, opened 2018-08-11
[NEW] Alternate Start -- New Beginnings

MacSplody:
Group
- Move to alt start group.

Bash tags
- Remove bad tag 'Sound'
- Add 'Scripts' Tags for vmad change.

Cleaning info
- Clean plug-in.

PR #288, opened 2018-08-12
Merge Updates - ETAC, Lighting Mods, ASLAL: NB & Verdant OCS

MacSplody:
[New] Expanded Towns and Cities (#282)
Add more incompatible messages for lighting mods (#280)
[NEW] Alternate Start -- New Beginnings (#285)
Flag latest Verdant OCS version as being Incompatible with Open Cities. (#287)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -339,7 +339,7 @@ plugins:
         nav: 2
     clean:
       - crc: 0xc18ee18c # v1.59.39.0 - Auto Clean
-        util: SSEEdit v3.2.1
+        util: 'SSEEdit v3.2.1'
   - name: 'Dawnguard.esm'
     group: *dlcGroup
     after:
@@ -408,9 +408,9 @@ plugins:
         nav: 57
     clean:
       - crc: 0x6061bcff # v1.59.39.0 - Auto Clean Second pass
-        util: SSEEdit v3.2.1
+        util: 'SSEEdit v3.2.1'
       - crc: 0x141be1e5 # v1.59.39.0 - Auto + Manual Cleaning
-        util: SSEEdit v3.2.1
+        util: 'SSEEdit v3.2.1'
   - name: 'Hearthfires.esm'
     group: *dlcGroup
     after:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -4190,6 +4212,315 @@ plugins:
...
+    tag:
+      - C.Location
+    clean:
+      - crc: 0xC0118C3B # v0.3.4
+        util: SSEEdit v3.2.2
+  - name: 'ETaC - Complete LoS Patch.esp'
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/13552']
+    msg:
+      - type: say
+        content: 'Plug-ins Masters require sorting.'
```
Freso: I'd say "Plugin’s". We generally use "plugin" without the dash elsewhere in the documentation, and the apostrophe is needed to denote genitive… otherwise we're talking about multiple plugins and then the sentence kind of doesn't make sense anymore. :)

masterlist.yaml
```diff
@@ -4190,6 +4212,315 @@ plugins:
...
+    tag:
+      - C.Location
+    clean:
+      - crc: 0x7E3B8E1B # v0.3.4
+        util: SSEEdit v3.2.2
+  - name: 'ETaC - Darkwater Crossing LoS Patch.esp'
+    url: ['https://www.nexusmods.com/skyrimspecialedition/mods/13552']
+    msg:
+      - type: say
+        content: 'Plug-ins Masters require sorting.'
```
Freso: If this is repeated, it seems like it should be made an anchor in `commons`.

Freso (changes requested):
Most of the changes here I don't really have an idea about, since I don't know the plugins involved at all, but the logic itself generally seems sane.

Only think I'd want to see here is a small wording (or rather grammar) change, and maybe having the involved `msg` moved up to the `commons` section.

masterlist.yaml
MacSplody: Okay I cleaned up the grammar, not my strong suit and added the message to `commons`

MacSplody:
Okay I cleaned up the grammar, not my strong suit and added the message to `commons`

masterlist.yaml
MacSplody: Done.

Freso (approved):
Looks good to me now, but again, I can't comment much on how it'll affect the game, since I haven't done Skyrim since before SSE was a thing. Syntactically it looks fine though. 👍

PR #305, opened 2018-08-20
Rebase Staging Commits To v0.13

MacSplody:


masterlist.yaml
```diff
@@ -6,6 +6,9 @@ common:
   - &SkyUI
     name: 'SkyUI_SE.esp'
     display: '[SkyUI](https://www.nexusmods.com/skyrimspecialedition/mods/12604)'
+  - &FNIS
+    name: 'tools/GenerateFNIS_for_Users/GenerateFNISforUsers.exe'
+    display: '[Fores New Idles in Skyrim SE](https://www.nexusmods.com/skyrimspecialedition/mods/3038)'
```
MacSplody, line 11: I had initially planned to use it with [XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988) but one of its options does not require FNIS to function.
So I had to switch from using only:
```yaml
    req:
      - name: *FNIS
```
to a message instead:
```yaml
    msg:
      - type: error
        content: 'You have installed [XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988/) features which require [Fores New Idles in Skyrim SE](https://www.nexusmods.com/skyrimspecialedition/mods/3038) to function correctly, but it is not installed.'
        condition: 'file("meshes/actors/character/character assets/skeleton.hkx") and not file("tools/GenerateFNIS_for_Users/GenerateFNISforUsers.exe")'
```
I can see a use for this `common` entry with later additions but it is currently not used.

masterlist.yaml
```diff
@@ -976,6 +1017,21 @@ plugins:
...
       - 'Relationship Dialogue Overhaul.esp'
+  - name: 'XPMSE.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/1988/'
+        name: 'XP32 Maximum Skeleton Special Extended - XPMSSE on Nexus Mods'
+    msg:
+      - <<: *doNotClean
+      - <<: *requireSKYUI
+        subs: [ '[XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988/)' ]
+        condition: 'file("scripts/XPMSEMCM.pex") and not active("SkyUI_SE.esp")'
```
MacSplody: Due to the set up of this mods Fomod, there were several scenarios where SkyUI would be required.
As this script was always present when SkyUI was required, it seemed like the easiest way to check.

masterlist.yaml
Freso: You can also do req entries with conditions, so you could have the req active only when the skeleton mesh is present. (Written on mobile, so sorry for terseness and lack of examples. :))

masterlist.yaml
```diff
@@ -957,6 +976,28 @@ plugins:
...
+    clean:
+      - crc: 0xEC081487 # v0.2.3
+        util: 'SSEEdit v3.2.40'
+  - name: 'RaceMenuPlugin.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/19080/' ]
+    req:
+      - 'RaceMenu.esp'
+    msg:
+      - type: say
+        content: 'This plugin is optional.'
```
MacSplody, line 997: Not really important, but might be of use to some users with a high plug-in count.

masterlist.yaml
```diff
@@ -976,6 +1017,21 @@ plugins:
...
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/1988/'
+        name: 'XP32 Maximum Skeleton Special Extended - XPMSSE on Nexus Mods'
+    msg:
+      - <<: *doNotClean
+      - <<: *requireSKYUI
+        subs: [ '[XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988/)' ]
+        condition: 'file("scripts/XPMSEMCM.pex") and not active("SkyUI_SE.esp")'
+      - type: error
+        content: 'You have installed [XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988/) features which require [RaceMenu Special Edition](https://www.nexusmods.com/skyrimspecialedition/mods/19080/) to function correctly, but it is not installed.'
+        condition: 'file("scripts/RaceMenuPluginXPMSE.pex") and not active("RaceMenu.esp")'
```
MacSplody: Same as above, several install scenarios & it was the most reliable way I could think of checking correctly.

masterlist.yaml
```diff
@@ -976,6 +1017,21 @@ plugins:
...
+      - type: error
+        content: 'You have installed [XP32 Maximum Skeleton Special Extended](https://www.nexusmods.com/skyrimspecialedition/mods/1988/) features which require [Fores New Idles in Skyrim SE](https://www.nexusmods.com/skyrimspecialedition/mods/3038) to function correctly, but it is not installed.'
+        condition: 'file("meshes/actors/character/character assets/skeleton.hkx") and not file("tools/GenerateFNIS_for_Users/GenerateFNISforUsers.exe")'
```
MacSplody: For this scenario, there's one install option which doesn't add the animation files `skeleton.hkx` and ` skeleton_female.hkx.`. Those animations require FNIS `GenerateFNISforUsers.exe` to work correctly.
If another mod where to add those files, then it would flag. But in that case I think they're going to need the version of XP32 Maximum Skeleton with those animations anyway, so this shouldn't be an issue.

masterlist.yaml
MacSplody: hmm I tried, oh well, guess I did something wrong. And no worries terseness never bothered me.

masterlist.yaml
```diff
@@ -2043,6 +2125,37 @@ plugins:
...
+      - 'Dolomite Weathers.esp'
+      - 'NAT.esp'
+    tag:
+      - Graphics
+    msg:
+      - <<: *hascompatibilityIssues
+        subs:
+          - '[MajesticMountains](https://www.nexusmods.com/skyrimspecialedition/mods/11052/)'
+          - '[Moss Rocks](https://www.nexusmods.com/skyrimspecialedition/mods/8838)'
+          - 'Moss Rocks can be used with Majestic Mountains. But Majestic Mountains meshes without snow will not be used. [Source](https://forums.nexusmods.com/index.php?/topic/5838597-majestic-mountains/?p=52271673)'
```
MacSplody, line 2138: The language on this might need some work.

masterlist.yaml
```diff
@@ -2139,6 +2252,9 @@ plugins:
     after:
       - 'SkyrimIsWindy.esp'
       - 'SRG Enhanced Trees Activator.esp'
+  - name: 'TerrainLodRedone.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/9135/' ]
+    group: *earlyLoadersGroup
```
MacSplody, line 2255: Set group to early loaders. Must load high to prevent it overwritting mods which add their own LOD such as Open Cities.
[Source](https://forums.nexusmods.com/index.php?/topic/5486077-lodterrain-lod-redone/?p=52022082 ) & Mod [description page]( https://www.nexusmods.com/skyrimspecialedition/mods/9135/ )

MacSplody:
I'll leave this open for a few hours in case anything should be changed or if anyone has any questions.

masterlist.yaml
Freso: Okay, so now I'm not on mobile. ;)
```yaml
    req:
      - name: *FNIS
```
Won't work because `&FNIS` already includes the `name`. What that above essentially is saying is:
```yaml
    req:
      - name:
        - name: 'tools/GenerateFNIS_for_Users/GenerateFNISforUsers.exe'
          display: '[Fores New Idles in Skyrim SE](https://www.nexusmods.com/skyrimspecialedition/mods/3038)'
```
which is not valid LOOT masterlist syntax. :)

Using
```yaml
    req:
      - *FNIS
```
Or
```yaml
    req:
      - <<: *FNIS
        condition: 'file("meshes/actors/character/character assets/skeleton.hkx")'
```
(Or even
```yaml
    req:
      - <<: *FNIS
        display: '[Fores New Idles in Skyrim SE](https://www.nexusmods.com/skyrimspecialedition/mods/3038) (when using this specific skeleton)'
        condition: 'file("meshes/actors/character/character assets/skeleton.hkx")'
```
) should work. :)

MacSplody: Well crap, thanks :)

Issue #336, opened 2018-09-28
JKs Skyrim_No Snow Under the Roof_Patch.esp


Teabag86:
Requested LOAD order
NSUTR_groundsnow_mesh_fixes.esp

JKs Skyrim_No Snow Under the Roof_Patch.esp OR
JKs Skyrim_Dawnstar_NSUTR_Patch.esp OR
JKs Skyrim_Dawnstar_BOYD_NSUTR_Patch.esp

The JKS patches need to load after the NSUTR mesh fixes.
There are only 3 records in conflict.
I should also note that only 1 of the last 3 files should be used, but I think you already have this covered. 

NSUTR mesh fixes https://www.nexusmods.com/skyrimspecialedition/mods/15064
JKS Patches https://www.nexusmods.com/skyrimspecialedition/mods/17026

Thanks

MacSplody:
@Teabag86 All done.

PR #340, opened 2018-09-28
Add JK's Skyrim - No Snow Under the Roof Patches

MacSplody:
Load after
- No Snow Under the roof - Hall of the Vigilant fix plugin NSUTR_groundsnow_mesh_fixes.esp

Add Message
- Warn that only one of the patches should be used.

Add cleaning info


MacSplody:
closes #336

PR #371, opened 2018-10-06
v1.5.53 Updates

LostDragonist:
Cleaned with SSEEdit v3.2.72 with simple records enabled

Actually tested this time so fingers crossed.


MacSplody (approved):
Looks good to me. Thanks for the contribution.

PR #377, opened 2018-10-08
Cleanup Has Wild Edits message

MacSplody:
- Remove plugin from Subs
- Remove Wild edit plugins from clean data.


masterlist.yaml
```diff
@@ -402,16 +400,10 @@ plugins:
         udr: 0
         nav: 57
     clean:
-      - crc: 0x6061bcff # v1.5.39.0 - Auto Clean Second pass
-        util: 'SSEEdit v3.2.1'
```
MacSplody, line 406: Removed CRC for plugins with wild edits. Because I think marking them as clean & also having wild edits is sending mixed messages.

PR #379, opened 2018-10-08
Modern Brawl Bug Fix Check Refactoring

MacSplody:


masterlist.yaml
```diff
@@ -93,6 +93,11 @@ common:
         text: 'Slet. Allerede inkluderet i %1%.'
       - lang: ja
         text: '削除してください。既に%1%に含まれています。'
+  - &includesMBBF
+    type: say
+    content:
+      - lang: en
+        text: '[ Modern Brawl Bug Fix ](https://www.nexusmods.com/skyrimspecialedition/mods/1473/) is included with this mod.'
```
MacSplody, line 100: Removed conditional, as it's useful for users to know which mods have this script.
Message type `say` is also not as intrusive as an `error`.

PR #385, opened 2018-10-15
Patch 7

Graybark:


Graybark:
Spent over an hour comparing formats and I have no idea why it keeps failing.

masterlist.yaml
```diff
@@ -5909,6 +5909,77 @@ plugins:
...
+        condition: 'active("Prometheus_No_snow_Under_the_roof.esp") and not active("Winterhold Restored - NSUTR Patch.esp")'
+      - <<: *hasPluginPatch
+        subs:
+          - 'Windyridge - Player Home In Winterhold'
+          - '[Winterhold Restored - Windyridge Patch](https://www.nexusmods.com/skyrimspecialedition/mods/12811/)'
+        condition: 'active("windyridge.esp") and not active("Winterhold Restored - Windyridge Patch.esp")'
+    tag:
+      - C.Location
+    clean:
+      - crc: E649AE58
```
```yaml
    clean:
      - crc: 0xE649AE58 # v1.9
        util: SSEEdit v3.2.77
```
Missing `0x` prefix for `CRC`.

masterlist.yaml
```diff
@@ -5909,6 +5909,77 @@ plugins:
...
+      - 'sm_winterholdgate.esp'
+      - 'Thunderchild - Epic Shout Package.esp'
+      - 'TKAA.esp'
+    msg:
+      - type: warn
+        condition: 'active("sm_jobasha_bookshop.esp")'
+        content: 'Jobasha''s Books SE is only compatible with its Winterhold Restored version.'
+        subs:
+          - 'Jobasha''s Books SE'
+          - '[Jobasha''s Books and Winterhold Restored Patch](https://www.nexusmods.com/skyrimspecialedition/mods/13285/)'
```
MacSplody: I would replace this with an incompatible message which only displays if the patch is not active.

```yaml
    inc:
      - 'ExpandedWinterholdRuins.esp'
      - 'Less desolate Winterhold.esp'
      - 'QaxeWinterholdRebuild.esp'
      - name: 'sm_jobasha_bookshop.esp'
        display: '[ Jobasha''s Books SE ](https://www.nexusmods.com/skyrimspecialedition/mods/13285/)'
        condition: 'active("sm_jobasha_bookshop.esp") and not active("sm_jobasha_WHrestore_patch.esp")'

```

masterlist.yaml
```diff
@@ -5909,6 +5909,77 @@ plugins:
...
+      - 'Immersive Citizens - AI Overhaul.esp'
+      - 'MagicalCollegeofWinterhold.esp'
+      - 'Skyrim Better Roads - Winterhold.esp'
+      - 'sm_winterholdgate.esp'
+      - 'Thunderchild - Epic Shout Package.esp'
+      - 'TKAA.esp'
+    msg:
+      - type: warn
+        condition: 'active("sm_jobasha_bookshop.esp")'
+        content: 'Jobasha''s Books SE is only compatible with its Winterhold Restored version.'
```
MacSplody: Mod specific messages should be placed in an entry for `sm_jobasha_bookshop.esp`.

'Jobasha''s Books SE is only compatible with its Winterhold Restored version.'

masterlist.yaml
Graybark: This is actually not a patch but a replacement version. You only use one of the plugins. Both are full versions of the mod. I wanted to put this there as a warning in case the user has downloaded the non-Winterhold Restored version.

masterlist.yaml
Graybark: I mean I could change it to say that Winterhold Restored is only compatible with the Winterhold Restored version of Jobasha's Books. Not sure I understand what you mean.

masterlist.yaml
MacSplody: So `sm_jobasha_WHrestore_patch.esp` is a replacer?

Okay well just set `sm_jobasha_bookshop.esp` as incompatible.

MacSplody: You can have the message, but I'll change it slightly, one sec.
```yaml
    msg:
      - type: warn
        content: 'Use the Winterhold Restored version of Jobasha''s Books SE instead: %1%'
        subs:
          - '[Jobasha''s Books and WinterHold Restored Patch] (https://www.nexusmods.com/skyrimspecialedition/mods/13285/)'
        condition: 'active("sm_jobasha_bookshop.esp")'
```

masterlist.yaml
```diff
@@ -5937,12 +5938,6 @@ plugins:
       - 'Thunderchild - Epic Shout Package.esp'
       - 'TKAA.esp'
     msg:
-      - type: warn
-        condition: 'active("sm_jobasha_bookshop.esp")'
-        content: 'Jobasha''s Books SE is only compatible with its Winterhold Restored version.'
-        subs:
-          - 'Jobasha''s Books SE'
-          - '[Jobasha''s Books and Winterhold Restored Patch](https://www.nexusmods.com/skyrimspecialedition/mods/13285/)'
```
MacSplody: This should work, if you're using `subs` the there needs to be a ref `%1%' to consume it.
```yaml
    msg:
      - type: warn
        content: 'Use the Winterhold Restored version of Jobasha''s Books SE instead: %1%'
        subs:
          - '[Jobasha''s Books and WinterHold Restored Patch] (https://www.nexusmods.com/skyrimspecialedition/mods/13285/)'
        condition: 'active("sm_jobasha_bookshop.esp")'
```

MacSplody (approved):
Perfect. Thanks for the contribution.

PR #392, opened 2018-10-20
 Add Quick Clean message for latest xEdit.

MacSplody:
I've added a seperate message for Quick cleaning rather than replacing manual cleaning completely.
So that it can be phased in to get users used to the change.
Plus the older cleaning information needs to be updated as it might be incorrect for Quick Clean.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -125,9 +125,11 @@ common:
     util: '[SSEEdit Experimental](https://github.com/TES5Edit/TES5Edit/releases)'
     info: |+
 
-          > **Quick Clean** is a xEdit startup option which will automatically clean a selected plugin.  
-          > A guide for running **Quick Clean** can be found [here](https://tes5edit.github.io/docs/5-mod-cleaning-and-error-checking.html#Apendix-C).  
-          > The latest version of [SSEEdit Experimental](https://github.com/TES5Edit/TES5Edit/releases) is required to use **Quick Clean**.  
+          >**Quick Clean** is a xEdit startup option which will automatically clean a selected plugin.
+          >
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -121,6 +121,14 @@ common:
   - &dirtyPlugin
     util: '[SSEEdit](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
     info: 'A cleaning guide is available [here](https://www.creationkit.com/index.php?title=TES5Edit_Cleaning_Guide_-_TES5Edit).'
+  - &quickClean
+    util: '[SSEEdit Experimental](https://github.com/TES5Edit/TES5Edit/releases)'
+    info: |+
+
+          > **Quick Clean** is a xEdit startup option which will automatically clean a selected plugin.
```
MacSplody: Not exactly sure about using a double space to force a newline on these block qoutes.

```yaml
    info: |

      >**Quick Clean** is a xEdit startup option which will automatically clean a selected plugin.
      >
      >A guide for running **Quick Clean** can be found [here](https://tes5edit.github.io/docs/5-mod-cleaning-and-error-checking.html#Apendix-C). 
      >
      >The latest version of [SSEEdit Experimental](https://github.com/TES5Edit/TES5Edit/releases) is required to use **Quick Clean**.
```
I could use that instead.

MacSplody: It was mainly to give it some space and make it clearer to the reader.
image

image

Ortham (approved):
The blockquote is a bit :/ but if it gives you the formatting you want, it is what it is.

MacSplody:
It is a bit of an eyesore, I'll review it at some point to see if I can come up with something better.

PR #393, opened 2018-10-22
Particle Patch for ENB SSE

Infernio:
- Appropriate Bash Tags added
- Clean CRC added
- Load After USSEP added (required for a few records and several meshes if you choose to pack the patch's loose files into a BSA)

Already tested locally, works as expected.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1083,6 +1083,16 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/2357/'
         name: 'Enhanced Blood Textures SE on Nexus Mods'
     group: *overhaulsGroup
+  - name: 'Particle Patch for ENB SSE.esp'
+    url:
+      - link: 'http://enbseries.enbdev.com/forum/viewtopic.php?t=1499'
+        name: 'Particle Patch for ENB SSE on the ENB Forums'
+    tag:
+      - C.Climate
+      - C.ImageSpace
+    clean:
+      - crc: 0xD0C086FD
+        util: 'SSEEdit v3.2.83 EXPERIMENTAL'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1083,6 +1083,18 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/2357/'
         name: 'Enhanced Blood Textures SE on Nexus Mods'
     group: *overhaulsGroup
+  - name: 'Particle Patch for ENB SSE.esp'
+    url:
+      - link: 'http://enbseries.enbdev.com/forum/viewtopic.php?t=1499'
+        name: 'Particle Patch for ENB SSE on the ENB Forums'
+    after:
+      - 'Unofficial Skyrim Special Edition Patch.esp'
```
MacSplody: It might be correct but It's not needed, `ESM` plugins can't sort after non `ESM` flagged plugins.

masterlist.yaml
```diff
@@ -1083,6 +1083,18 @@ plugins:
...
+    tag:
+      - C.Climate
+      - C.ImageSpace
```
MacSplody, line 1092: Yeah those are fine.

MacSplody (changes requested):
You can remove the load after rule, as it's not required.

masterlist.yaml
Infernio: My particle patch _is_ ESM and ESL tagged, because of ESL-related load order concerns (I need to patch some parts of the patch later in the load order).

If we don't want to worry about that possibility, I'll remove that `after:`.

Infernio:
Load after rule is gone.

MacSplody (approved):
👍

PR #394, opened 2018-10-22
Replaced all instances of tags: with tag:

Infernio:
There were several instances of `tags:` in the masterlist, which is not actually recognized by LOOT - those should be `tag:`.

Affected Entries:
 - Better Dynamic Snow.esp
 - Better Dynamic Snow - NSUTR Patch.esp
 - Improved Traps.esp
 - MLU - AAE Ultimate Edition.esp
 - Circlets and Masks with all Robes and Hoods.esp

Already tested locally, works as expected.


MacSplody (approved):
Well spotted, thanks.

PR #403, opened 2018-10-27
Several improvements to the german localization in the masterlist

Infernio:
There was some incorrect grammar and strange phrasing.
I also unified the style, which is now a consistent formal 'Sie'.


MacSplody (approved):
👍

PR #405, opened 2018-10-27
Enhanced Blood Textures and SkyBirds Compatibility Patch

Infernio:
- Added a warning telling users to install the patch
- Made the patches and script range modifiers incompatible with each other
- Made the splatter size modifiers incompatible with each other

Tested locally, works as expected.


MacSplody (approved):
👍 Thanks for the contribution.

Issue #292, opened 2018-08-16
xEdit Quick Auto Clean Reports


MacSplody:
Checked if added to Master List. 
Strikeout if addition is not required.
Links are to reports below.

---
- [x]  [DLC](https://github.com/loot/skyrimse/issues/292#issuecomment-423271332)
---
  - [x]   [Eli's Breezehome Overhaul](https://github.com/loot/skyrimse/issues/292#issuecomment-423272240)
---
- [x] [The Haven From The Cold And Dark 2.0](https://github.com/loot/skyrimse/issues/292#issuecomment-423273456)
---
- [x] [Interesting NPCs](https://github.com/loot/skyrimse/issues/292#issuecomment-423278686)
---
- [x] [Holds The City Overhaul](https://github.com/loot/skyrimse/issues/292#issuecomment-423283352)
---
- [x] [Beyond Skyrim - Bruma SE](https://github.com/loot/skyrimse/issues/292#issuecomment-423286985)
---
- [x] [Rigmor of Bruma](https://github.com/loot/skyrimse/issues/292#issuecomment-423288726)
---
- [x] [Palaces and Castles Enhanced SSE](https://github.com/loot/skyrimse/issues/292#issuecomment-423295128)
---
- [x] [VIGILANT SE](https://github.com/loot/skyrimse/issues/292#issuecomment-423303685)
---
- [x] [M'rissi's Tails of Troubles SE](https://github.com/loot/skyrimse/issues/292#issuecomment-423305898/)
---
- [x] [Hammet's Dungeon Packs for SE](https://github.com/loot/skyrimse/issues/292#issuecomment-423581425)

MacSplody:
```
 - name: 'Update.esm'
    dirty:
      - <<: *dirtyPlugin
        crc: 0xE95F6A98
        util: 'SSEEdit v3.2.67 EXPERIMENTAL'
        itm: 256
        udr: 91
        nav: 2
    dirty:
      - <<: *dirtyPlugin
        crc: 0xA208CF56
        util: 'SSEEdit v3.2.67 EXPERIMENTAL'
        nav: 2

  - name: 'Dawnguard.esm'
    dirty:
      - <<: *dirtyPlugin
        crc: 0xCFBEE94B
        util: 'SSEEdit v3.2.68 EXPERIMENTAL'
        itm: 724
        udr: 82
        nav: 57
    dirty:
      - <<: *dirtyPlugin
        crc: 0xC1E55A15
[...]

PR #406, opened 2018-10-27
Quick Auto Clean Updates

Infernio:
Added cleaning info for all mods listed in #292 

Went with 'atomic commits' for this one, but you can squash them if you like


MacSplody (approved):
👍 Thanks

PR #407, opened 2018-10-27
Refactoring: Removed all unnecessary itm:, udr: and nav: entries

Infernio:
They're redundant and made searching through the list for entries with deleted navmeshes difficult.

While I'm at it, I also marked the three mods with deleted navmeshes on the list as dirty:
- The Haven From The Cold And Dark.esp
- Distinct Interiors.esp
- Eli_Breezehome.esp

That said, as the TODO comments in the PR imply, we might want to wait with this until `*noAutomaticFix` is added.


MacSplody (approved):
👍

PR #433, opened 2018-11-03
Distinct Interiors.esp: Change the dirty markers to reqManualFix

Infernio:
This one seems to have been missed in the recent 'deleted navmeshes' rework.
Marked as requiring a manual fix both before and after cleaning.


Infernio:
Hold on, investigating this further right now
This record seems to not have been updated for a bit

Infernio:
Redone in #434

PR #434, opened 2018-11-03
Distinct Interiors: Properly handle the deleted navmeshes

Infernio:
Well, this turned out to be a much bigger endeavor than I thought...

As it turns out, after a discussion with Elminster on Discord, the one deleted navmesh in `Distinct Interiors.esp` and `Distinct Interiors - Player Homes.esp` is forwarded from `HearthFires.esm`, making it a sort of wild edit that doesn't really cause harm besides xEdit complaining about it.

This PR adds instructions on how to manually clean the affected plugins and also adds clean CRCs for afterwards.


MacSplody (approved):
👍

PR #445, opened 2018-11-07
Early groups

MacSplody:


Ortham:
One potential problem with this that I forgot about when you asked before: things like the Unofficial Patches should be loading before pretty much everything else, does this change not mix them up with everything else that needs to load before default but still after them?

MacSplody:
@Ortham Yes they will be in the same group, but I still have testing to do to make sure everything is behaving as expected. I can move USSEP to `Light Masters` group & maybe rename it if I have to add a lot of `after` rules.

MacSplody:
@Ortham #455 Will Keep USSEP first in most cases.

PR #476, opened 2018-11-23
Update RS Children + CRF Patch ESP

ann4belle:
CRF modpage no longer includes patch. Swapped message to patch3rdParty and added link.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5224,9 +5224,11 @@ plugins:
       - 'UnlimitedBookshelves.esp'
       - 'WhiterunHoldForest.esp'
     msg:
-      - <<: *patchProvided
-        subs: [ 'RS Children Overhaul' ]
-        condition: 'active("RSkyrimChildren.esm") and not active("RSChildren - CRF Patch.esp")'
+      - <<: *patch3rdParty
+        subs: 
+          - 'RS Children Overhaul' 
+          - '[ RS Children Overhaul Cutting Room Floor Patch ](https://www.nexusmods.com/skyrimspecialedition/mods/8753)'
+        condition: 'active("RSkyrimChildren.esm") and not active("CRF_RSChildrenPatch.esp")'
     tag:
       - Actors.ACBS
       - Actors.AIData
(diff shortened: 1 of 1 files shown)


MacSplody (changes requested):
There is also another version included with [RS Children Patch Compendium](https://www.nexusmods.com/skyrimspecialedition/mods/13409) `RSC CRF Patch.esp`

```yaml
      - <<: *patch3rdParty
        subs: 
          - 'RS Children Overhaul' 
          - '[ RS Children Overhaul Cutting Room Floor Patch ](https://www.nexusmods.com/skyrimspecialedition/mods/8753)'
        condition: 'active("RSkyrimChildren.esm") and not active("CRF_RSChildrenPatch.esp") and not file("RSC CRF Patch.esp")'
      - <<: *patch3rdParty
        subs: 
          - 'RS Children Overhaul' 
          - '[ RS Children Patch Compendium ](https://www.nexusmods.com/skyrimspecialedition/mods/13409)'
        condition: 'active("RSkyrimChildren.esm") and not active("RSC CRF Patch.esp") and not file("CRF_RSChildrenPatch.esp")'
```
Above will cover either patch being used, messages will only display if neither patch is present, and only for one mod if patch is present but deactivated.

ann4belle:
```
- <<: *patch3rdParty
  subs: 
    - 'RS Children Overhaul' 
    - '[ RS Children Patch Compendium ](https://www.nexusmods.com/skyrimspecialedition/mods/13409)'
  condition: 'active("RSkyrimChildren.esm") and not active("RSC CRF Patch.esp") and not active("CRF_RSChildrenPatch.esp")'
```
Would it not be better to do this, and link the Compendium for user convenience, while still having support for the other patch?

MacSplody:
I've tried that approach before, it just causes confusion when they have a patch installed but the linked patch for the warning message is different.
There is no harm in having both, you'll see both links if no patch is installed.
Only one message if either is installed but is not active.
So they will see both and can decide which they want to use. You could swap the order to have the compendium message at the top.

MacSplody (approved):

MacSplody:
👍 Thanks for the contribution @Not-A-Walnut

PR #509, opened 2018-12-08
Several miscellaneous updates for follower mods

Infernio:
- Inigo: Cleaned the mod with xEdit 4.0.0, which finds several ITMs
- Lucien: Same as above, also finds several ITMs
- Lucien: Added info for all provided patches

Also moved Less Dragons above 'KrittaKittyHorsesForSSE', since that's correct and I was editing in the vicinity.


Infernio:
Whoops, need to test this locally. One second.

Infernio:
Redone in #510.

PR #635, opened 2019-02-16
Bash Tags: C.GridFlags has been renamed to C.ForceHideLand

Infernio:
The upcoming `C.GridFlags` tag has been renamed to `C.ForceHideLand` to better match the xEdit / CK name for the field it patches - see https://github.com/wrye-bash/wrye-bash/commit/86c4beab4603e40d12bc991acccb7e0479d8e59b.
This doesn't really affect anything, since that tag isn't actually used in the masterlist yet.


MacSplody:
@Infernio Thanks

PR #641, opened 2019-02-27
Missing Encounter Zones Fixed - Add Info

lucible:
I'd like to add LOOT rules to the main masterlist for the mod Missing Encounter Zones Fixed; I've checked the syntax etc, on my local repository and all appears to be working as it should!

Thank you!


masterlist.yaml
```diff
@@ -9774,3 +9774,54 @@ plugins:
         util: 'SSEEdit v4.0.1n RC4 for 4.0.2'
       - crc: 0xDA25CBFC
         util: 'SSEEdit v4.0.1n RC4 for 4.0.2'
+  - name: 'MissingEZs_AllExteriors.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/23609'
+        name: 'Missing Encounter Zones Fixed on Nexus Mods'
+    group: *highPriorityGroup
```
MacSplody: If you're setting a group, then you'll need to know if it should load after mods or before mods in that group.
It's to avoid sorting errors.

Enhanced Lighting for ENB (ELE)
ELFXEnhancer
ELFX - Hardcore
Luminosity - Lightweight Lighting
Luminosity Lighting Overhaul
Alternate Start - Live Another Life
Open Cities Skyrim
Immersive Citizens - AI Overhaul

MacSplody: Enhanced Lighting for ENB (ELE)
image

masterlist.yaml
```diff
@@ -9774,3 +9774,54 @@ plugins:
...
+    msg:
+      - <<: *patchProvided
+        subs: [ 'Realistic Water Two' ]
+        condition: 'active("RealisticWaterTwo.esp") and not active("MZEF_ExteriorsRW2Patch.esp")'
+  - name: 'MZEF_ExteriorsRW2Patch.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/23609'
+        name: 'Missing Encounter Zones Fixed on Nexus Mods'
+    after:
+      - 'RealisticWaterTwo.esp'
```
MacSplody: You don't need to add an `after` if the plugin is a master.
image

masterlist.yaml
```diff
@@ -9774,3 +9774,54 @@ plugins:
...
+      - 'RelightingSkyrim_SSE.esp'
+      - 'Populated Dungns Caves Ruins Legendary.esp'
+    tag:
+      - C.Encounter
+      - C.Location
+    msg:
+      - <<: *patchProvided
+        subs: [ 'Realistic Water Two' ]
+        condition: 'active("RealisticWaterTwo.esp") and not active("MZEF_ExteriorsRW2Patch.esp")'
+  - name: 'MZEF_ExteriorsRW2Patch.esp'
```
MacSplody, line 9795: Missing Change from RWT
image

masterlist.yaml
MacSplody: One conflict with Open Cities
image

lucible: Yes, it should load before ELE and ELFX - those mods need patches for this mod. It should load after Immersive Citizens, though.

It will have minor conflicts with mods such as Open Cities; those are generally harmless / can easily be patched by mod users in xEdit since it's so heavily dependent on load order.

masterlist.yaml
lucible: good to know! should I change that and re-commit/pull?

masterlist.yaml
lucible: how did that sneak past me? thanks! 😅

masterlist.yaml
MacSplody: If it should be after Immersive Citizens then, 
put it in the same group as ICAIO, and add a load after as well.
 group: *lowPriorityGroup

masterlist.yaml
MacSplody: Also 
image

masterlist.yaml
MacSplody: Yes you can remove that after rule for the patch.
Just make another commit to this pull request.

masterlist.yaml
MacSplody: SSEEdit Modgroups make it much easier to spot those.

MacSplody (changes requested):
Okay so changes are
- Drop the afters from the patches.
- Place the main plugin in the Low Priority Group
- Set the main plugin to load after Immersive Citizens.

masterlist.yaml
lucible: that's bizarre, I'm not seeing any of these conflicts with RW2 in SSEdit :S

MacSplody: I'll see if I have an Older version.

MacSplody (approved):
Looks good now, Thanks.

PR #680, opened 2019-05-20
Add Smilodon and load order for Immersive Movement

sean-kang:
1. Smilodon is a spin-off of Wildcat and its metadata are almost same as Wildcat.
2. Immersive Movement should be loaded before Wildcat, Smilodon and Mortal Enemies as instructed in the FOMOD installer of Immersive Movement.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6479,26 +6479,6 @@ plugins:
       # version: 1.7
       - crc: 0x725F1E2E
         util: 'SSEEdit v3.2.1'
-  - name: 'Smilodon - Combat of Skyrim.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/2824/' ]
-    inc:
-      - 'Wildcat - Combat of Skyrim.esp'
-    after:
-      - 'Animal Tweaks.esp'
-      - 'Immersive Movement.esp'
-      - 'Immersive NPC in the dark.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -6477,6 +6479,26 @@ plugins:
       # version: 1.7
       - crc: 0x725F1E2E
         util: 'SSEEdit v3.2.1'
+  - name: 'Smilodon - Combat of Skyrim.esp'
```
MacSplody: > Smilodon is a spin-off of Wildcat and its metadata are almost same as Wildcat.

In that case, I'd place `Smilodon - Combat of Skyrim.esp` directly after the entry for `Wildcat - Combat of Skyrim.esp`, and before the `Wildcat - Realistic Damage Plugin.esp`.

The order of entries in the master-list has no effect. 
It's perfectly fine to ignore the order. if a more appropriate placement is available.

A regex for shared metadata could then be added later without too much noise.
```yaml
  - name: '(Wildcat|Smilodon) - Combat of Skyrim.esp'
```
Smilodon also has a Realistic Damage Plugin, so I think they could also share metadata with a regex.

sean-kang: I created a new entry `(Smilodon|Wildcat) - Combat of Skyrim\.esp` to share metadata. Please confirm if it looks all right. Thanks.

MacSplody (approved):
Looks good, Thanks.

PR #684, opened 2019-06-03
Add Apocalypse compatibility patch for Vokrii

sean-kang:
A patch for the new perks mod Vokrii is provided by Apocalypse.


MacSplody:
👍 Thanks.

PR #685, opened 2019-06-03
Add patch message for RS Children

sean-kang:
A compatibility patch for Beyond Skyrim - Bruma is included in RS Children.


MacSplody:
@sean-kang Thanks.

PR #686, opened 2019-06-05
Add compatibility patches

sean-kang:
1. Add CACO and Wintersun compatibility patch
2. Add Summermyst and WACCF compatibility patch


MacSplody:
@sean-kang Thanks.

PR #693, opened 2019-06-11
Add messages for RS Children and CACO compatibility patches

sean-kang:


MacSplody (commented):
Changes are fine, I'd split that up into at least two commits though.
Commits with a `&` or an `and` usually indicate seperate commits should be used.

https://www.freshconsulting.com/atomic-commits/

sean-kang:
I thought it would be squashed like it was before. That's okay. Here are new split commits.

MacSplody:
> 
> 
> I thought it would be squashed like it was before. That's okay. Here are new split commits.

No worries, and thanks.
I try to avoid squashing where possible. I'll generally only squash to tidy up the commits if there are corrections, and all the commits are to the same mod. Make as many commits as you feel you need, within reason of course.

MacSplody (approved):
Perfect, thanks.

PR #695, opened 2019-06-15
Add messages for iNeed compatibility patches

sean-kang:
Notes: 'iNeed CACO Patch.esp' is different from kryptopyr's script patch. This patch distributes CACO foods into iNeed food categories of light, medium and heavy.


MacSplody:
@sean-kang Thanks.

PR #718, opened 2019-06-24
Load RDO after Auri

Xahtax:
Load RDO after Auri


MacSplody:
@Xahtax Thanks. 👍 

Just make sure to change the summary in Github Desktop when making a commit.

image

image

PR #739, opened 2019-08-07
Adding "Improved Rain and Fog for Cathedral Weathers" + patch

Nawor3565:


masterlist.yaml
```diff
@@ -12960,3 +12960,13 @@ plugins:
     after:
       - 'TeenDolls - Lewd.esp'
       - 'TeenDolls - Physics(Official).esp'
+  - name: 'Obsidian In Cathedral.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/27831' ]
+    group: *Late Fixes & Changes
```
MacSplody: That group will put it very near the end, is there some reason for this?
What is it conflicting with that it needs to be so late?

Nawor3565: You're right, it doesn't really have to be that late, I suppose default is fine.

MacSplody: If you have some idea of specific mods it should be after or before that would help.

Nawor3565: As long as the first file is after Cathedral Weathers (which I don't believe is on the masterlist) and the patch is after the first file, it should be fine.

MacSplody: Okay I just checked, and only this rule is required.
```yaml
  - name: 'DarkerNights-ObsidianInCathedral.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/27831' ]
    after:
      - 'Obsidian In Cathedral.esp'
```

`Obsidian In Cathedral.esp` has `Cathedral Weather.esp` as a master.

image

So LOOT will already sort `Obsidian In Cathedral.esp` after `Cathedral Weather.esp` 

You can remove the followeing in that case.
```yaml
  - name: 'Obsidian In Cathedral.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/27831' ]
    after:
      - 'Cathedral Weathers.esp'
```

Nawor3565: Alright, made the changes. Thanks for your help

MacSplody (approved):
Perfect and thanks @Nawor3565

PR #743, opened 2019-08-11
added russian localization for messages where it was not added

TokcDK:
added russian localization for messages where it was not added (atleast main of them which on top of masterlist)


MacSplody:
@TokcDK Thanks.

PR #759, opened 2019-09-02
Placing Sounds of Skyrim/Cathedral Weathers Patch below Improved Cathedral Weathers

Nawor3565:
Based on a post on the Nexus, the game will not play nicely unless the mods are in this order.


MacSplody:
> Based on a post on the Nexus, the game will not play nicely unless the mods are in this order.

image

Well there no conflicts between them, but it shouldn't cause any issues to add the after.
It can be reviewed if it ever does cause problems.
It's a shame they didn't give some more details like their load order to test.
I'd bet there's something else causing the issue, and the after just happens to solve it.

MacSplody (approved):

Issue #691, opened 2019-06-09
Unofficial Skyrim Special Edition Patch 4.1.8a and package of Aventus Aretino 


DFelten:
Currently there is a problem with the Unofficial Skyrim Patch and some children mods:

https://forums.nexusmods.com/index.php?showtopic=4978860/#entry70583578

If one of these mods is active together with Unofficial Skyrim Special Edition Patch 4.1.8a, the game does not load anymore:

* TK Children
* RS Children
* Simple Children
* Killable Children
* Axe's RS Children Mod
* Improved Adoptions
* Immersive Weapon Integration

MacSplody:
@DFelten Thanks, just have to decide on the best way to handle that issue.

> ALERT FROM UNOFFICIAL PATCH PROJECT TEAM: It has been determined that the AI package USKPAventusAretinoSleepOrphanage [PACK:05000808] on Actor AventusAretino "Aventus Aretino" [NPC_:00014132] is redundant and has been removed from Aventus and deleted. However USSEP compatibility solutions with various mods altering Aventus still refer to the now nonexistent AI pack. This results on a infinite-loading screen on the main menu when starting the game.
> 
> Users, if you are familiar with xEdit, remove the unresolved reference error that now exists in this mod's compatibility solution with USSEP. Mod author, it is recommended you do the same and re-release for those not fluent in xEdit. 

AventusAretino 00014132
PKID - Package
[05000808] < Error: Could not be resolved >

Modules affected:
[Axe's RS Children](https://www.nexusmods.com/skyrimspecialedition/mods/16044)
`aAxeRSCtweaks.esp` (26F29854)

[Killable Children](https://www.nexusmods.com/skyrimspecialedition/mods/1108)
`Killable Children.esp` (C38F2891)

[RS Children Overhaul](https://www.nexusmods.com/skyrimspecialedition/mods/2650)
`RSChildren.esp` (6407E29C) Non Playable - USSEP
`RSChildren.esp` (2CC8F8D2) Playable - USSEP

[Simple Children](https://www.nexusmods.com/skyrimspecialedition/mods/22789)
`FacegenForKids.esp` (DB8F0931)

[TKChildren SE](https://www.nexusmods.com/skyrimspecialedition/mods/5916)
`TKChildren.esp` (8B45CAFD) English Version
`TKChildren.esp` (B5A9AA32) Japanese Version

Fixed Mods
Improved Adoptions USSEP fix.esp (EE0A02C2)
[...]

PR #795, opened 2019-09-21
Remove USSEP dependency on RSkyrimChildren.esm

skarkkai:
If RSkyrimChildren.esm is loaded before USSEP, SSE will spin on the load screen forever.

Got this tip from https://www.reddit.com/r/skyrimmods/comments/c3l4qx/stuck_on_main_screen_with_the_spinning_loading/ and found the fix to work for me.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1533,7 +1533,6 @@ plugins:
     after:
       - name: 'Lanterns Of Skyrim - All In One - Main.esm'
         condition: 'checksum("Lanterns Of Skyrim - All In One - Main.esm", EB94B5F8)'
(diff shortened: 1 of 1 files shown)


MacSplody:
@skarkkai Changing the load order between the two might be getting you to the main menu, but it's not fixing the actual issue. 

An unresolved reference error in Aventus's AI package is what's causing the game to remain on the loading screen. There's an issue open for it here. #691 

If you are familiar with xEdit, you need to remove the unresolved reference in `00014132` _AventusAretino_ Packages from `RSChildren.esp`.

image

image

image

image

skarkkai:
@CptMcSplody Thank you!

PR #798, opened 2019-09-22
Load Realistic Room Rental after Economy Overhaul

sean-kang:
If [Realistic Room Rental](https://www.nexusmods.com/skyrimspecialedition/mods/24859) does not load after [Economy Overhaul](https://www.nexusmods.com/skyrimspecialedition/mods/9542), its scripts cannot use varied inn costs in the MCM settings for each inn. It will instead use Economy Overhaul's fixed 25 septims for every inn.


MacSplody:
> It will instead use Economy Overhaul's fixed 25 septims for every inn.

👌 

image

MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #862, opened 2019-10-15
Added some swedish translations

simoneddeland:


pStyl3:
Thanks for the translations! There are two issues that prevents us from merging your changes right now. 

First, you should rebase your fork, so that we are on equal grounds. Your version of the masterlist currently doesn't include [this](https://github.com/loot/skyrimse/commit/764b10cd6545b28e6a0ae64daaf8e967eaf8b0c7) for example. Basically, we set an order for all the translations, which goes as follows:

```yaml
00 English
01 Czech
02 Danish
03 German
04 Spanish
05 Finnish
06 French
07 Japanese
08 Korean
09 Polish
10 Portuguese
11 Brazilian Portuguese
12 Russian
13 Swedish
14 Chinese

00 en
01 cs
02 da
03 de
04 es
05 fi
06 fr
07 ja
08 ko
09 pl
10 pt
11 pt_BR
12 ru
13 sv
14 zh_CN
```
Your Swedish translation should therefore be last if no Chinese translation is given. If it is, then before that (Chinese translations should always be absolute last).

Second, you've set your PR so that it should merge in the `loc-fi` branch. The correct one would be the `v0.14` branch.

It probably would be best to close this PR, rebase your fork, assure that everything is correct for your translations, and create a new PR to be merged into the `v0.14` branch.

simoneddeland:
I had some problems rebasing the new commit to my fork so I simple re-forked and added my translations to the correct branch, see [this](https://github.com/loot/skyrimse/pull/863) PR

PR #866, opened 2019-10-17
Add Economy Overhaul patches and set load order

sean-kang:
Add Economy Overhaul compatibility patches and set load order of the patches. The installer says Vokrii and Ordinator patches should be loaded after Morrowloot patch.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6666,8 +6666,14 @@ plugins:
         subs: [ 'Trade and Barter' ]
         condition: 'active("Trade & Barter.esp") and not active("EconomyOverhaul CV - TradeandBarter Patch.esp")'
       - <<: *patchIncluded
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -6693,6 +6699,9 @@ plugins:
       # version: 2.1
       - crc: 0xBBA1C2CA
         util: 'SSEEdit v4.0.2'
+  - name: 'EconomyOverhaul CV - (Vokrii|Ordinator) Patch.esp'
+    after:
+      - 'EconomyOverhaul CV - MorrowLootUltimate Patch.esp'
```
MacSplody, line 6704: Group order should cover this after, but there's no harm in having it since it's correct.

MacSplody (approved):

PR #873, opened 2019-10-21
Load Landscape and Water Fix after Landscape Fixes for Grass Mods

sean-kang:
The author of Landscape and Water Fix said, "If you have Landscape Fixes for Grass Mods﻿ installed, place the ESP version below it.".

Source: https://www.nexusmods.com/skyrimspecialedition/mods/26138?tab=description

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5381,6 +5381,8 @@ plugins:
 
   - name: 'Landscape and Water Fixes.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/26138/' ]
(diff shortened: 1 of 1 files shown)


MacSplody:
@sean-kang That rule can't be added at the moment. Anyone using the `ESM` flagged version would receive a sorting error. It will have to wait until https://github.com/loot/libloot/issues/60 is implemented and released.

Default sorting will place `Landscape and Water Fixes.esp` after `Landscape Fixes For Grass Mods.esp`, so LOOT is already sorting it correctly.

image

image

PR #879, opened 2019-10-30
Add Survival Mode patches for some popular mods

ketsuban:
Nexus user Conner has made [some patches for popular mods](https://www.nexusmods.com/skyrimspecialedition/mods/19154) to get them to work nicely with Survival Mode, but LOOT doesn't know about these patches to recommend them.

I think I did this right - I'm not unfamiliar with YAML, but I'm far from an expert. I didn't do all the patches, because I looked at the file listing for Zim's Immersive Artifacts and laziness descended on me like a storm cloud. Sorry.


MacSplody (approved):

MacSplody:
@ketsuban Thanks.

PR #967, opened 2020-01-01
Add ENB Light.esp

mrudat:
Add dependencies for ENB Light. Possibly needs bash tags, also has a UPF patcher, which might require a message.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2576,6 +2576,13 @@ plugins:
       - name: 'RLO - Interiors.esp'
         condition: 'not file("Realistic Lighting Overhaul - Major City Interiors.esp")'
 
+  - name: 'ENB Light.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/22574' ]
+    after:
+      - 'ELE_SSE.esp'
+    req:
+      - 'Particle Patch for ENB SSE.esp'
+
   - name: 'ImprovedHearthfireLighting.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -2576,6 +2576,13 @@ plugins:
       - name: 'RLO - Interiors.esp'
         condition: 'not file("Realistic Lighting Overhaul - Major City Interiors.esp")'
 
+  - name: 'ENB Light.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/22574' ]
+    after:
+      - 'ELE_SSE.esp'
```
MacSplody, line 2582: It would still require a patch, but I've no objections to the after.
Bash for SSE doesn't handle light records, so no tags are required at the moment.

masterlist.yaml
```diff
@@ -2576,6 +2576,13 @@ plugins:
...
+    req:
+      - 'Particle Patch for ENB SSE.esp'
```
MacSplody, line 2584: My only concern with the req is the plugin `Particle Patch for ENB SSE.esp` has been superceded by [ENB Helper](https://www.nexusmods.com/skyrimspecialedition/mods/23174). the meshes and texture from particle patch can still be used with ENB helper. But it's not major as users can still disable `Particle Patch for ENB SSE.esp` if using ENB Helper and the requirement will be fufilled.

MacSplody (approved):
Looks good to me.

MacSplody:
@mrudat Thanks 👍

PR #1001, opened 2020-01-19
Add sort masters message to Colored Map Markers

Chef0000:
Added info regarding sorting MAST records for Colored Map Markers for SSE.


MacSplody (approved):

MacSplody:
@Chef0000 Thanks.

PR #1010, opened 2020-01-26
Load Diverse Skyrim after Kireina to avoid black face problem

sean-kang:
Diverse Skyrim needs to be loaded after Kireina that changes the default facegen of NPCs in order to fix black face problem.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5172,6 +5172,9 @@ plugins:
 
   - name: 'DIVERSE SKYRIM.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/7707/' ]
+    after:
+      - 'Kireina SSE.esl'
+      - 'Kireina SSE.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -5172,6 +5172,9 @@ plugins:
 
   - name: 'DIVERSE SKYRIM.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/7707/' ]
+    after:
+      - 'Kireina SSE.esl'
```
MacSplody, line 5176: This after shouldn't be required as `DIVERSE SKYRIM.esp` will sort after the `.esl` version.
Maybe if they ESM flagged `DIVERSE SKYRIM.esp`, can't imagine why they would though.
It's not going to cause any issue either way so it can be left in the pr.

MacSplody (approved):
@sean-kang Thanks.

PR #1029, opened 2020-02-08
Load Skyrim Revamped after Lore Worthy Bandits

sean-kang:
Skyrim Revamped rebalances loots and it should not be overwritten by Lore Worthy Bandits. It is the same reason as why Lore Worthy Bandits' author recommends Morrowloot to be loaded after his mod.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1043, opened 2020-03-08
Add VR perk extender requirement for Economy Overhaul complete version

sean-kang:
Skyrim VR CTD'd when it ran Economy Overhaul Complete version without VR perk extender because the complete version added many perks to speech tree.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1071, opened 2020-03-24
Add load order for Bugfix Compilation

sean-kang:
Bugfix Compilation should follow the load order of Hunters Not Bandits because the changes from the mod is included in Bugfix Compilation. Since the regular version of Hunters Not Bandits is included in Bugfix Compilation, it should be overwritten by RDO version of Hunters Not Bandits.

Load order: RDO -> Bugfix Compilation(Regular version of Hunters Not Bandits included) -> Hunters Not Bandits(RDO version)

Reference: https://www.nexusmods.com/skyrimspecialedition/mods/33449?tab=description

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1570,6 +1570,10 @@ plugins:
         util: 'SSEEdit v4.0.3'
 
 ###### Fixes ######
+  - name: 'BugfixCompilation.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/33449/' ]
+    group: *fixesGroup
+
   - name: 'Butterflies.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/29434/' ]
     group: *fixesGroup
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1570,6 +1570,10 @@ plugins:
         util: 'SSEEdit v4.0.3'
 
 ###### Fixes ######
+  - name: 'BugfixCompilation.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/33449/' ]
+    after: [ 'Relationship Dialogue Overhaul.esp' ]
```
MacSplody: From [Bugfix Compilation](https://www.nexusmods.com/skyrimspecialedition/mods/33449) description page.

image

> Compatibility
> 
> This should be compatible with just about everything. It is a low priority mod, meaning that it should be at the very top of your load order. Let other mods overwrite it.

image

> Hunters Not Bandits by raccoondance
> 
> The regular version is included. In case you're using RDO, just let it overwrite this compilation or download "Hunters Not Bandits for RDO" and place it after both mods.

Based on that and the overall content of the mod.
I would place it early as suggested in the `Fixes & Resources` group.

```suggestion
    group: *fixesGroup
```

sean-kang:
Updated as McSplody suggested.

MacSplody:
@sean-kang Thanks.

PR #1077, opened 2020-04-04
Fix yaml parsing error

sean-kang:
Here is a fix. Sorry about this.


MacSplody:
@sean-kang No worries, I didn't see it either. You can test the changes locally just to be 100% sure it works.

PR #1078, opened 2020-04-06
Add Interesting NPCs patch messages

sean-kang:


sean-kang:
It seems that TravisCI failed to send a success message back to GitHub. If you see the details in TravisCI, you will see the task had successfully completed.

PR #1088, opened 2020-04-10
Add Convenient Carriages patch messages

sean-kang:


sean-kang:
This is a partial work because there are too many patches included in the mod. I wish someone else who uses city overhaul mods to add rest of them.

MacSplody (approved):

PR #1092, opened 2020-04-12
Add Mysticism CRF patch message and load order

sean-kang:
1. Add [Mysticism CRF patch](https://www.nexusmods.com/skyrimspecialedition/mods/34010) message
2. Load order: Mysticism CRF patch should be loaded after [Flesh FX](https://www.nexusmods.com/skyrimspecialedition/mods/27456) since Mysticism CRF patch already has the same shader record changes and its magic effect changes shouldn't be overwritten by vanilla values in Flesh FX.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1102, opened 2020-04-16
Add CLLF load order and patch message

sean-kang:
Although the author of CLLF claims that the mod is fully compatible with WACCF, it actually is not. To avoid some part of CLLF's leveled lists being overwritten by WACCF, it has to be loaded after WACCF. It is the same reason why MLU has to be loaded after WACCF.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1103, opened 2020-04-17
Add SkyVRaan Shimmering VR Waters load order

### LINKED ISSUE #1087: SkyVRaan Shimmering Waters Plugin Order
I've created a mod called SkyVRaan for Skyrim VR. In VR, the water reflections have been disabled, which generally looks very bad. I have a plugin named Cubemapenabler.esp that defines a cubemap for exterior water the same way it can be defined for an interior cell. I do this by assigning XWEM parameter for exterior worldspaces and every cell in those worldspaces.

Unfortunately, when assigning cubemaps in this way, you are limited to two water types the same way that an interior cell would be. If there are more than two watertypes defined, it causes instability and crashes. For this reason, cubemapenabler.esp also changes all watertypes to defaultwater. 

currently cubemapenabler is sorted very high up in the load order by loot. It should be loaded after bash and smash patches, but before DynDOLOD or Occlusion. My instructions on my mod page are to put cubemapenabler after Bash and Smash, and then run DynDOLOD and generate xLODGEN Occlusion with cubemapenabler enabled.

If Bash or Smash patches are loaded after cubemapenabler, or run with cubemapenabler active, then they will revert my watertype changes and cause crashes.

Please let me know if you would like more information. My mod page is:

https://www.nexusmods.com/skyrimspecialedition/mods/30571

sean-kang:
According to the author of the mod, cubemap enabler should be at the very end of load order to overwrite water type data. However, it could be before or after DynDOLOD depending on how DynDOLOD.esp was built.

Water grass addon should be loaded after Landscape Fixes for grass mod.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3014,6 +3014,12 @@ plugins:
         util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 5
 
+  - name: 'CubemapEnabler.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/30571/' ]
+    group: *lateChangesGroup
+  - name: 'SkyVRaan-WaterGrass.esp'
+    after: [ 'Landscape Fixes For Grass Mods.esp' ]
+
   - name: 'RealisticWaterTwo.*\.esp'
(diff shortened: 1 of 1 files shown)


MacSplody:
@sean-kang Thanks.

PR #1104, opened 2020-04-18
Add Containers and Leveled Lists Fixes load order

sean-kang:
WACCF, ACE, WAF, Lore worthy bandits should not overwrite leveled list changes of CLLF. It is the same rule as MLU load order.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6910,6 +6910,20 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/18571/' ]
     after: [ 'Requiem.esp' ]
 
+  - name: 'Containers and Leveled Lists Fixes.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/26575/' ]
+    group: *economyGroup
+    after:
+      - 'LoreWorthyBandits.esp'
+      - 'MLU.esp'
+      - 'Weapon AF.esp'
+      - 'WACCF_Armor and Clothing Extension.esp'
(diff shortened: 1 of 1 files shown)


MacSplody:
> It is the same rule as MLU load order.

It could be put into the economy group.
And It can still be used with MLU, according to the author, as long as it's after MLU.
If it was added to the economy group an after would have to be added to `Trade & Barter.esp`.
To place it after `Containers and Leveled Lists Fixes.esp`, otherwise there would be sorting errors from patches.

sean-kang:
I made changes as you suggested. Please check if it looks all right.

masterlist.yaml
```diff
@@ -6910,6 +6910,19 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/18571/' ]
     after: [ 'Requiem.esp' ]
 
+  - name: 'Containers and Leveled Lists Fixes.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/26575/' ]
+    group: *economyGroup
+    after:
+      - 'LoreWorthyBandits.esp'
+      - 'Weapon AF.esp'
```
MacSplody: Looks fine, it needs an after for Morrowloot Ultimate as well.

```suggestion
      - 'LoreWorthyBandits.esp'
      - 'MLU.esp'
      - 'Weapon AF.esp'
```

sean-kang:
Updated. Thanks.

MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1110, opened 2020-04-19
Fix for Trade & Barter Patch Name

PenguinMancer:
The patch for Trade & Barter and Keld-Nar was named incorrectly.
The mod author appears to have renamed the file after a version change. 
This will match the new name.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -7385,7 +7385,7 @@ plugins:
         subs:
           - 'Keld-Nar'
           - '[Trade and Barter - Patches](https://www.nexusmods.com/skyrimspecialedition/mods/23220/)'
-        condition: 'active("Keld-Nar.esp") and not active("Trade and Barter - Keld-Nar Patch.esp")'
+        condition: 'active("Keld-Nar.esp") and not active("Keld-Nar - Trade and Barter Patch.esp")'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -7385,7 +7385,7 @@ plugins:
         subs:
           - 'Keld-Nar'
           - '[Trade and Barter - Patches](https://www.nexusmods.com/skyrimspecialedition/mods/23220/)'
-        condition: 'active("Keld-Nar.esp") and not active("Trade and Barter - Keld-Nar Patch.esp")'
+        condition: 'active("Keld-Nar.esp") and not active("Keld-Nar - Trade and Barter Patch.esp")'
```
MacSplody: Probably best to keep both names, just for anyone that renamed the file.

```suggestion
        condition: 'active("Keld-Nar.esp") and not active("Trade and Barter - Keld-Nar Patch.esp") and not active("Keld-Nar - Trade and Barter Patch.esp")'
```

PenguinMancer: Oh yeah that's a good point...

MacSplody (approved):

MacSplody:
@PenguinMancer Thanks.

PR #1119, opened 2020-04-27
Update Skyrim Revamped load order

sean-kang:
WACCF, ACE, WAF should not overwrite the leveled list changes of Skyrim Revamped.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1120, opened 2020-04-27
Sort Armor Rating Redux after Smilodon

sean-kang:
Armor rating data from Armor Rating Redux should not be overwritten by Smilodon.


MacSplody (approved):

MacSplody:
@sean-kang Thanks.

PR #1130, opened 2020-05-02
Recommend BP tweak over Feminine Females

Infernio:
Applies to the entire load order, so tell BP users that they can just use the tweak instead.
Same with the really old 'OGA Fix' mod.

Not sure about that big ol regex, maybe just `Mors Feminine Female.*\.esp`?

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -527,6 +527,16 @@ common:
       - lang: sv
         text: 'Denna plugins master behöver sorteras'
 
+  - &useBashTweakInstead
+    type: say
+    content:
+      - lang: en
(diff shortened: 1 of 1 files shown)


MacSplody:
> Not sure about that big ol regex, maybe just `Mors Feminine Female.*\.esp`?

It works fine, so I don't have an issue with it.
But if you prefer it that way then feel free to change it.

Infernio:
Changed it to `.*`, the other one just felt too ugly to me 😛

MacSplody (approved):

MacSplody:
@Infernio Thanks.

PR #1176, opened 2020-06-10
Adding XP32 XPMSSE Skeleton ITM Patch

cecell:
I"m about to publish this patch, it needs to be added to the lateLoadersGroup before I click publish though.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -16632,3 +16632,9 @@ plugins:
     clean:
       - crc: 0xA1170B2D
         util: 'SSEEdit v4.0.3'
+
+  ### XP32 XPMSSE Skeleton ITM Patch ###
+  - name: 'XP32 XPMSE Skeleton - ITM Protection Agency.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -16632,3 +16632,9 @@ plugins:
     clean:
       - crc: 0xA1170B2D
         util: 'SSEEdit v4.0.3'
+
+  ### XP32 XPMSSE Skeleton ITM Patch ###
+  - name: 'XP32 XPMSE Skeleton - ITM Protection Agency.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/37142' ]
+    group: *lateLoadersGroup
+    after: [ 'XPMSE.esp' ]
```
MacSplody, line 16640: If `XPMSE.esp` is a master file for `XP32 XPMSE Skeleton - ITM Protection Agency.esp`, then you won't need this after.

cecell: It isn't because I'm trying to stay super generic.  You'll see in the mod description.  It's a bit dumb but I seriously have no idea how people will react to the smallest things...

PR #1222, opened 2020-08-16
Tuned rules for patch to new version of patched mod

jgostling:
Added version checking of Majestic Mountains. Patch no longer needed as of v 3.10.


MacSplody (approved):

MacSplody:
@jgostling Thanks.

PR #1232, opened 2020-08-20
Add bash tag to Convenient Horses

sibir-ine:
added Text tag to address conflict with USSEP (since it has Text tag too and overwrites an edit by CH)


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1233, opened 2020-08-20
Add tags and load-after rules for Realistic Needs and Diseases

sibir-ine:
- Hunterborn and WACCF undo effects changes that are best addressed by loading RNAD after

- tags are to forward leveled list, graphics, and keyword changes undone by mods like USSEP

- this is all done with [RNAD 2](https://www.nexusmods.com/skyrimspecialedition/mods/23799) and [HB + MCM](https://www.nexusmods.com/skyrimspecialedition/mods/17993), but these changes should benefit other versions as well


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1236, opened 2020-08-20
Edit entry for Toccata as Elisif

sibir-ine:
- should load after Pan's NPC overhaul as that mod also overhauls Elisif, making this one redundant otherwise
- updated the cleaning info at the same time, no change in checksum

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -4212,10 +4212,11 @@ plugins:
 
   - name: 'Toccata as Elisif.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12713/' ]
+    after: [ 'PAN_NPCs.esp' ]
     clean:
       # version: 1.4.2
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -4212,10 +4212,11 @@ plugins:
 
   - name: 'Toccata as Elisif.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12713/' ]
+    after: [ 'PAN_NPCs.esp' ]
```
MacSplody, line 4215: It should already do this, but there's no harm in having it added.

MacSplody (approved):

masterlist.yaml
sibir-ine: ah, sorry! i remember having that custom rule for ages in LE as well because i don't think it always did

PR #1238, opened 2020-08-21
Update CBBE WACCF patch message and add CBBE ACE patch message

sibir-ine:
- updated the CBBE WACCF patch message to use a common one with a better description and more translations and to go away when patch is installed
- added CBBE ACE patch message with same functionality
- made a mistake with comments that was addressed

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3715,10 +3715,8 @@ plugins:
           - '[Weapons Armor Clothing and Clutter Fixes - CBBE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/19176/)'
         condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp") and not file("CalienteTools/BodySlide/SliderSets/WACCF.osp")'
     clean:
-        #1.6.1
       - crc: 0xD600DFE6
         util: 'SSEEdit v4.0.3'
-        #1.6.1 ESL
       - crc: 0x7187204B
         util: 'SSEEdit v4.0.3'
   - name: 'RaceMenuMorphsCBBE.esp'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3714,11 +3714,14 @@ plugins:
           - 'Weapons Armor Clothing and Clutter Fixes'
           - '[Weapons Armor Clothing and Clutter Fixes - CBBE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/19176/)'
         condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp") and not file("CalienteTools/BodySlide/SliderSets/WACCF.osp")'
+      - <<: *patch3rdParty
+        subs:
+          - 'Armor and Clothing Extension'
+          - '[Weapons Armor Clothing and Clutter Fixes - CBBE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/19176/)'
+        condition: 'active("WACCF_Armor and Clothing Extension.esp") and not file("CalienteTools/BodySlide/SliderSets/ACE.osp")'
     clean:
-        #1.6.1
       - crc: 0xD600DFE6
         util: 'SSEEdit v4.0.3'
-        #1.6.1 ESL
       - crc: 0x7187204B
         util: 'SSEEdit v4.0.3'
   - name: 'RaceMenuMorphsCBBE.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -3709,19 +3709,16 @@ plugins:
...
         subs:
           - 'Weapons Armor Clothing and Clutter Fixes'
           - '[Weapons Armor Clothing and Clutter Fixes - CBBE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/19176/)'
-        condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp")'
+        condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp") and not file("CalienteTools/BodySlide/SliderSets/WACCF.osp")'
     clean:
+        #1.6.1
       - crc: 0xD600DFE6
         util: 'SSEEdit v4.0.3'
+        #1.6.1 ESL
```
MacSplody: I used to do it this way, but I find it's just as easy to leave the information in the commit/pr message.
Like this https://github.com/loot/skyrimse/pull/1234/commits/c46014e287586701e0a7b50cb7e4b0a31ec8f2f8, it can be useful in some cases.

```suggestion
      - crc: 0xD600DFE6
        util: 'SSEEdit v4.0.3'
```

masterlist.yaml
```diff
@@ -3709,19 +3709,16 @@ plugins:
       - <<: *alsoUseX
         subs: [ '[Unofficial Skyrim Special Edition Patch](https://www.nexusmods.com/skyrimspecialedition/mods/266/)' ]
         condition: 'not active("Unofficial Skyrim Special Edition Patch.esp")'
-      - type: say
-        content:
-          - lang: en
-            text: 'You seem to be using %1%. A compatibility patch is available here %2%.'
-          - lang: de
-            text: 'Sie scheinen %1% zu nutzen. Ein Kompatibilitätspatch ist hier verfügbar: %2%'
+      - <<: *patch3rdParty
```
MacSplody, line 3712: > 'You seem to be using **%1%**, but you have not enabled a compatibility patch for this mod. A third party patch is available here: %2%'

The old wording was because it's not a plugin and can't be enabled.

masterlist.yaml
```diff
@@ -3709,19 +3709,16 @@ plugins:
...
         subs:
           - 'Weapons Armor Clothing and Clutter Fixes'
           - '[Weapons Armor Clothing and Clutter Fixes - CBBE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/19176/)'
-        condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp")'
+        condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp") and not file("CalienteTools/BodySlide/SliderSets/WACCF.osp")'
```
MacSplody, line 3716: That should do fine for verifying and could also be used for checking the ACE patch.
`ACE.osp` is installed by the patch fomod for [Armor and Clothing Extension](https://www.nexusmods.com/skyrimspecialedition/mods/19002).

masterlist.yaml
sibir-ine: got it, just saw it in another entry and figured i'd replicate it. i'll follow your method in the future

masterlist.yaml
sibir-ine: figured as much. i just figured that the patch3rdParty message didn't imply the patch was a plugin either, so it'd be better to take advantage of more translations and a slightly more descriptive message in combination with the condition to check if it's installed

masterlist.yaml
sibir-ine: didn't know that existed, i can implement a message for that patch as well if you'd like!

masterlist.yaml
MacSplody: Right, well It shouldn't cause too much confusion using the patch3rdParty message.

masterlist.yaml
MacSplody: Okay, you can commit the above suggested change directly. I would do it, but it messes up the pull request authors.
image

masterlist.yaml
MacSplody: Sure if you like, you could add it to this pull request or in a new one.

sibir-ine: just about to push it to this pull request

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1239, opened 2020-08-21
Add some more bash tags to mods

Infernio:
- XxAwesome_PotionsxX SSE (ugh)
- Guard Dialogue Overhaul
- Zim's Immersive Artifacts
- Dragon Bridge
- Spectral/Luminous Obsidian Weathers Patch
- ~~Landscape Fixes for Grass Mods - Cutting Room Floor Locations~~ (see discussion below)
- Vokriinator
- Heavy Armory

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -10346,6 +10346,7 @@ plugins:
     tag:
       - Delev
       - Invent
+      - Keywords
       - Names
       - Relev
       - Stats
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8030,6 +8030,7 @@ plugins:
       - 'Ambriel.esp'
       - 'Cutting Room Floor.esp'
       - 'Hothtrooper44_ArmorCompilation.esp'
+    tag: [ Keywords ]
     clean:
       # version: 1.4.2 Redux, True GDO Port
       - crc: 0x6ABC1129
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -3318,6 +3318,10 @@ plugins:
       - crc: 0x795E49B4
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Obsidian_TS_Patch_(Luminous|Spectral)\.esp'
```
MacSplody: There's an entry for those files at https://github.com/loot/skyrimse/blob/962439626a0b47dab2c16a246e453945a1e38b9a/masterlist.yaml#L3501.

masterlist.yaml
```diff
@@ -5438,6 +5442,7 @@ plugins:
       - crc: 0xA40E1836
         util: 'SSEEdit v4.0.3'
   - name: 'Landscape Fixes For Grass mods - Cutting Room Floor Locations.esp'
+    tag: [ C.Location ]
```
MacSplody: I didn't think that plugin made any changes to cell locations.

Infernio: `00009366` and `00009367`:

image

Probably not terribly important since this is a location conflict, which generally means one of the two mods *will* break.

masterlist.yaml
Infernio: Thanks, will move it there.

masterlist.yaml
MacSplody: Pretty sure that change is from an old version of Cutting Room Floor, 
I guess they never noticed the change to location.

Infernio: That's possible. I'll drop the tag then.

MacSplody (approved):
The rest looks fine.

MacSplody:
@Infernio Thanks.

PR #1241, opened 2020-08-22
Add/update various tags, cleaning info, and messages

sibir-ine:
- created entries for RN&D 2 - Frostfall patch, Qwinn's ACE - CRF patch, The Eyes of Beauty, and SEELODGen terrain plugin
- added tags for RN&D 2 - FF patch, Shor's Stone, Notice Board - DB patch, Qwinn's ACE - CRF patch, and WACCF
- added cleaning info for new entries, especially The Eyes of Beauty, verified/updated rest along the way
- added message to indicate SSELODGen terrain plugin is optional as it's only needed for generating terrain LOD but harmless if left in
- unsure where to put the SSELODGen plugin as it's a modding tool but not generated by the user
- a more ambitious change that can be left out if it's excessive: created a common message for optional plugins as a few plugins use it and more probably could. it would also be easier for translators and contributors

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -16870,6 +16870,11 @@ plugins:
       - <<: *alreadyInX
         subs: [ 'kryptopyr''s Patch Hub - CCOR & Cloaks of Skyrim Patch' ]
         condition: 'active("Cloaks_CCOR_Patch.esp")'
+  - name: 'Qw_ACE_CRF Patch.esp'
+    tag: [ Keywords ]
+    clean:
+      - crc: 0xE2B0065F
+        util: 'SSEEdit v4.0.3'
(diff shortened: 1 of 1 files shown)


MacSplody:
Okay most of the changes are fine, but I think they're too many unrelated changes for one pull request.
So these should be split up into seperate pull requests.
https://github.com/loot/skyrimse/pull/1241/commits/130acc409485ac5dfeb30299697372ce51697f47, could be one PR for example.
The idea would be to try and keep commits and pull requests as focused as possible.
They don't have to be perfect, just something to aim for.
https://www.freshconsulting.com/atomic-commits/, explains it better.

sibir-ine:
ah, sorry! got a little carried away. should i just close this pull request and separate them? i'll work on separating them tomorrow. feel free to close it if i don't respond

PR #1249, opened 2020-08-23
Create common optional message

sibir-ine:
- created common optional message as 3 entries use it and more probably could
- it would be easier for translators to translate it and for contributors to use it for other entries
- message updated in WACCF Bash plugin and RaceMenu plugin entries
- message also updated in SSELODGen Terrain plugin entry i added in another pull request that hasn't been merged yet. it will use it's own if this pull request is rejected


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1256, opened 2020-08-25
Add load rule to AI Overhaul

sibir-ine:
* AIO should load after Dawnstar [according to one of the authors](https://forums.nexusmods.com/index.php?/topic/7205116-ai-overhaul-sse/page-115#entry84067243) to address a few conflicts

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8003,6 +8003,7 @@ plugins:
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/21654'
         name: 'AI Overhaul SSE'
+    after: [ 'Dawnstar.esp' ]
     tag:
       - Actors.ACBS
       - Actors.AIData


MacSplody:
>     * AIO should load after Dawnstar [according to one of the authors](https://forums.nexusmods.com/index.php?/topic/7205116-ai-overhaul-sse/page-115#entry84067243) to address a few conflicts

The other author says to just leave it be.

image

Though I don't think @Arthmoor was shown the full conflicts.

image

It seems pretty minor to me with only three records overlapping, one could easily patch those conflicts.

image

image

sibir-ine:
the author of AIO has patches for some of Arthmoor's other town overhauls, so they could probably make a patch for Dawnstar as well. however, given that they saw the 2 main conflicts (the packages) in the thread i linked, that also makes me think that they prefer that AIO wins the conflicts outright

Arthmoor:
They may prefer that AIO win, but then my changes to the packages will be lost and the NPCs won't spend time in their homes at all. Since the only conflicts of interest anyone ever showed me were for the two AI packs, there was (and IMO remains) no logical reason to let AIO break those changes.

sibir-ine:
got it, i'll close the PR. as for a patch, i'm not familiar with AI package records. since these conflicts seem more severe than with your other town overhauls, would you be willing to lay out what changes need to be preserved so they or someone else could make a proper patch? from looking at the conflicts, i'd guess it's not as simple as "drag forward and drop a couple times"

PR #1261, opened 2020-08-28
Changes to DLC entries

sibir-ine:
* added Invent.Add and Invent.Change to Update.esm
  - list of relevant records: 000ABD9E, 000B2035, 00078C0C
* replaced deprecated Invent tags with appropriate new tags
* removed Relev tag from Dragonborn.esm
  - someone may have added it for this record, 00105242, but Relev doesn't forward the LVLD change
* inventory and leveled list tags were the only i considered
* updated wild edits guide URL
* removed unnecessary load rule
  - Hearthfire already loads after Dawnguard because of line 1141

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1154,9 +1154,7 @@ plugins:
         nav: 5
   - name: 'Dragonborn.esm'
     group: *dlcGroup
-    after:
-      - 'Dawnguard.esm'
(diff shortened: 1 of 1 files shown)


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1283, opened 2020-09-06
Update Skyrim Particle Patch for ENB

sibir-ine:
* Based on [this post](http://enbseries.enbdev.com/forum/viewtopic.php?f=6&t=1499&start=1420#p84446) by the author
* Plugin is only needed by people who use ENB but not SKSE, but ENB Helper is preferred
* Preferred alsoUseX over alreadyInOrFixedByX or useInstead as only the plugin is deprecated, not the resources
* Wanted to use deletePlugin, but the sentence about it being a requirement would be inaccurate
* Instead copied the contents of deletePlugin and removed the sentence about a requirement from each translation as well
  - pt_BR was unchanged as it already doesn't seem to mention a requirement


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1284, opened 2020-09-13
Update Skyrim Landscape and Water Fixes

sibir-ine:
* Added new checksum of updated SMIM-compatible mesh

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5467,7 +5467,7 @@ plugins:
         condition: 'active("RelightingSkyrim_SSE.esp") and not active("Landscape and Water Fixes - Patch - Relighting Skyrim.esp")'
       - <<: *patchIncluded
         subs: [ 'Static Mesh Improvement Mod' ]
-        condition: 'active("SMIM.*\.esp") and not checksum("meshes/architecture/farmhouse/walkway/walkwaycwall01.nif", E114830E)'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -5467,7 +5467,7 @@ plugins:
         condition: 'active("RelightingSkyrim_SSE.esp") and not active("Landscape and Water Fixes - Patch - Relighting Skyrim.esp")'
       - <<: *patchIncluded
         subs: [ 'Static Mesh Improvement Mod' ]
-        condition: 'active("SMIM.*\.esp") and not checksum("meshes/architecture/farmhouse/walkway/walkwaycwall01.nif", E114830E)'
+        condition: 'active("SMIM.*\.esp") and (not checksum("meshes/architecture/farmhouse/walkway/walkwaycwall01.nif", E114830E) and not checksum("meshes/architecture/farmhouse/walkway/walkwaycwall01.nif", 9016DCA9))'
```
MacSplody, line 5470: This is the downside of using CRC for file checks and why I try to avoid them.

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

masterlist.yaml
sibir-ine: Definitely, but the patch only consists of meshes. There are also 3 versions of the file (vanilla, SMIM, and Enhanced textures detail (UV-tweaks)), so just checking that the file isn't the one from SMIM isn't enough. I'll make sure to keep an eye on the mod in case it needs updates, though.

PR #1299, opened 2020-09-19
Add Serana Dialogue Add-On

sibir-ine:
* Added location for plugin and patches
* Incompatible with mods that make Serana marriable (e.g. [Marry Me Serana](https://www.nexusmods.com/skyrimspecialedition/mods/21938/)) as it has its own custom implementation
* Added [SDE](https://www.nexusmods.com/skyrimspecialedition/mods/16222/) recommendation as this mod was designed as a companion to it, but it isn't a hard requirement
* Added messages for some patches included on mod page
  - DX Crimson Armor is a bit of a mess as there are 2 versions ([UNP](https://www.nexusmods.com/skyrimspecialedition/mods/29451/) & [CBBE](https://www.nexusmods.com/skyrimspecialedition/mods/29197/)) with multiple versions of the plugin each (2 for UNP, 7 for CBBE). Figured a basic message would be enough, but I could list out all the checksums if that's preferred
  - There are more patches for mods I avoid, so I'd rather avoid potentially poorly implementing and supporting them
* Added Keywords tag
  - all included LCTN records
* Added cleaning info
  - 1.8.2

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8474,6 +8474,10 @@ plugins:
       - crc: 0xF3EF2798
         util: 'SSEEdit v4.0.2f'
 
+  - name: '(SeranaDialogAddon|SDA).*\.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/32161/' ]
+  - name: 'SeranaDialogAddon.esp'
+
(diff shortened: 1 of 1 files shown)


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1328, opened 2020-09-27
Triss Armor Retextured - WACCF Compatibility

sibir-ine:
* Sort WACCF after optional [TAR](https://www.nexusmods.com/skyrimspecialedition/mods/23745/) plugin to eliminate conflict

conflict
* Added message for [consistency patch](https://www.nexusmods.com/skyrimspecialedition/mods/27667/)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -7543,6 +7543,7 @@ plugins:
       - name: 'Guard Dialogue Overhaul.esp'
         condition: 'checksum("Guard Dialogue Overhaul.esp", 6ABC1129) or checksum("Guard Dialogue Overhaul.esp", 5CB27D7D) or checksum("Guard Dialogue Overhaul.esp", E248614B) or checksum("Guard Dialogue Overhaul.esp", E40EDE50) or checksum("Guard Dialogue Overhaul.esp", 7D16C40A)'
       - 'Immersive Sounds - Compendium.esp'
+      - 'TAR-GuildMaster1stPersonFix.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -7543,6 +7543,7 @@ plugins:
       - name: 'Guard Dialogue Overhaul.esp'
         condition: 'checksum("Guard Dialogue Overhaul.esp", 6ABC1129) or checksum("Guard Dialogue Overhaul.esp", 5CB27D7D) or checksum("Guard Dialogue Overhaul.esp", E248614B) or checksum("Guard Dialogue Overhaul.esp", E40EDE50) or checksum("Guard Dialogue Overhaul.esp", 7D16C40A)'
       - 'Immersive Sounds - Compendium.esp'
+      - 'TAR-GuildMaster1stPersonFix.esp'
```
MacSplody, line 7546: If the changes are included in WACCF, then maybe an `alreadyInOrFixedByX` message would be better?

sibir-ine: Ah, sorry, I should've gone into more detail as to why I opted for the load after rule over that. The plugin also edits the Thieves Guild leader gloves I assume to make them more consistent with the new model it replaces them with. Doesn't look like the biggest issue if the message is still preferred.

conflict

MacSplody: Okay, an after makes sense in that case.

MacSplody (approved):

PR #1341, opened 2020-10-03
Update tags for a few more plugins

sibir-ine:
Added:
- Update.esm
- Dragonborn.esm
- Arthmoor's Darkwater Crossing
- Arthmoor's Karthwasten
- Better Skill & Quest Book Names
- Castle Volkihar Rebuilt
- EFF
- Fall of Granite Hill
- Keep It Clean

Removed:
- Ars Metallica
- Arthmoor's Dragon Bridge
- ASLAL
- JK's Skyhaven Forge
- Notice Board - Better Solstheim Quests

Updated:
- Immersive Armor
- Improved Traps
- Trade & Barter

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12761,7 +12761,9 @@ plugins:
       - 'ViscousGrass.esp'
       - 'Immersive Citizens - AI Overhaul.esp'
       - 'Landscape Fixes For Grass Mods.esp'
-    tag: [ Names ]
+    tag:
+      - Keywords
+      - Names
     clean:
       # version: 2.0.6
       - crc: 0x9CCF0D4E
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -13137,6 +13137,7 @@ plugins:
     after:
       - 'HoldBorderBanners.esp'
       - 'Immersive Citizens - AI Overhaul.esp'
+    tag: [ Keywords ]
     clean:
       # version: 1.0.5
       - crc: 0x362EA12F
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12797,7 +12797,9 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/8683/' ]
     after: [ 'Landscape Fixes For Grass Mods.esp' ]
     inc: [ 'DragonBridgeBoardwalk.esp' ]
-    tag: [ Actors.AIPackages ]
+    tag:
+      - Actors.AIPackages
+      - -Invent.Remove
     clean:
       # version: 2.0.5
       - crc: 0x88C9CF62
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6725,6 +6725,9 @@ plugins:
     msg:
       - <<: *patch3rdParty_LotD
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_ArsMetallica_Patch.esp")'
+    tag:
+      - -Invent.Remove
+      - -Names
     clean:
       - crc: 0x4688FC6A
         util: 'SSEEdit v4.0.3'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6200,8 +6200,10 @@ plugins:
   - name: 'Improved Traps.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/17592/' ]
     tag:
-      - Graphics
-      - Stats
+      - name: SpellStats
+        condition: 'checksum("Improved Traps.esp", F718BAC1)'
+      - name: Stats
+        condition: 'checksum("Improved Traps.esp", F718BAC1)'
     clean:
       # version: 2.3 - Normal Edition
       - crc: 0xDAA5AC38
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1085,6 +1085,8 @@ plugins:
   - name: 'Update.esm'
     group: *dlcGroup
     tag:
+      - C.Encounter
```
MacSplody: You won't need that if USSEP is being used, or if the `Dawnguard.esm` wild edits have been cleaned.

masterlist.yaml
```diff
@@ -6721,6 +6727,9 @@ plugins:
     msg:
       - <<: *patch3rdParty_LotD
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_ArsMetallica_Patch.esp")'
+    tag:
+      - -Invent.Remove
```
MacSplody: It's an Invent tag in the plugin header.
Maybe remove that and replace it with Invent.Add.
```suggestion
      - -Invent
      - Invent.Add
```

masterlist.yaml
```diff
@@ -12788,7 +12800,9 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/8683/' ]
     after: [ 'Landscape Fixes For Grass Mods.esp' ]
     inc: [ 'DragonBridgeBoardwalk.esp' ]
-    tag: [ Actors.AIPackages ]
+    tag:
+      - Actors.AIPackages
+      - -Invent.Remove
```
MacSplody: Same suggestion as the Invent for Ars Metallica.

```suggestion
      - -Invent
      - Invent.Add
```

masterlist.yaml
sibir-ine: That's a better solution, thanks!

masterlist.yaml
```diff
@@ -10362,7 +10371,8 @@ plugins:
         condition: 'active("Complete Crafting Overhaul_Remastered.esp") and not active("Hothtrooper44_ArmorCompilation_CCOR_Patch.esp")'
     tag:
       - Delev
-      - Relev
```
MacSplody, line 10365: It was probably unintended behaviour in bash.
But having a Relev tag used to carry forward changes to Level List Flags.
I think that's why some mods like this and Trade & Barter had them included.

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

masterlist.yaml
sibir-ine: Yeah, I figured it had something to do with that. It's not mentioned in the documentation, so I assumed it didn't still do that. From testing with Trade & Barter, it looks like it still does. Not only that, but without the Relev tag, WB undoes changes to the flags even if it's the only plugin editing the leveled list.

conflict

I'll reassess my Relev removals and open another PR in a bit. I'll also bring it up in the WB discord as it looks like this is either all unintended behavior or there's only some unintended behavior and the docs need to be updated.

MacSplody: > I'll reassess my Relev removals and open another PR in a bit. I'll also bring it up in the WB discord as it looks like this is either all unintended behavior or there's only some unintended behavior and the docs need to be updated.

It would be simpler if the flags, chance and global were handled separately.
Adding Relev as a work around is going to affect the overall output of both Delev and Relev.

sibir-ine: Okay, I'll leave it as is and bring it up on the WB discord.

sibir-ine: From @Infernio:

> Yes, the leveled list patcher has really stupid flag handling logic. I want to rewrite the whole patcher eventually, so adding more tags onto the broken mess that it is doesn't make sense right now.

masterlist.yaml
MacSplody: It seems `-Invent` doesn't apply `-Invent.Remove` or `-Invent.Add`.
So the tags will still be `Invent.Add` & `Invent.Remove`.
We could ignore the `Invent` tag and treat it as only having `Invent.Add` & `Invent.Remove`.
But this tag is in the plugin header, and if it's removed, it would end up with no tags.
So we have to treat it as having the tags, and potentially not having the tags.
Suggesting removal of `Invent` doesn't really matter, as it won't have any effect in bash.

```yaml
    tag:
      - Invent.Add
      - -Invent.Remove
```

Infernio: That's because WB doesn't internally recognize `Invent` as a tag anymore. As soon as it sees any deprecated tags in a description etc., it replaces them with their updated versions - in this case both `Invent.Add` and `Invent.Remove`. So if the masterlist then tries to remove the `Invent` tag, that does nothing, because no `Invent` tag is applied to that plugin anymore (only the two Add/Remove tags).

MacSplody: Yeah, I just assumed at the time that the removal of deprecated tags was taken into account when updating.

masterlist.yaml
MacSplody: > > Yes, the leveled list patcher has really stupid flag handling logic. I want to rewrite the whole patcher eventually, so adding more tags onto the broken mess that it is doesn't make sense right now.

@sibir-ine @Infernio 
We can restore the `Relev` tags for now, with a comment to note why.
And review it if how flags are handled is ever changed in the leveled list patcher.

sibir-ine: Got it, I'll review my Relev and Invent changes and open a PR some time today.

PR #1342, opened 2020-10-03
Updates for Convenient Carriages & Unique Region Names

sibir-ine:
Before load after rule:
* 0201A521
conflict0

After load after rule:
* 00008E96
conflict1
* 0000B4B9
conflict2

* Updated locations for patches/add-ons for each mod while I was at it

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6450,8 +6450,9 @@ plugins:
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 773
 
-  - name: 'Unique Region Names.esp'
+  - name: '(Unique Region Names|URN.*).esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/11750/' ]
+  - name: 'Unique Region Names.esp'
     group: *lowPriorityGroup
     after:
       - name: 'Immersive Citizens - AI Overhaul.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -11873,6 +11874,8 @@ plugins:
       # ICAIO - Default
       - crc: 0x8EEAB63F
         util: 'SSEEdit v4.0.3'
+  - name: 'CC_Ferry_Patch.esp'
+    after: [ 'Unique Region Names.esp' ]
```
MacSplody, line 11878: That's fine for now, I'll probably put that activator in a separate plugin.

masterlist.yaml
```diff
@@ -6450,8 +6450,9 @@ plugins:
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 773
 
-  - name: 'Unique Region Names.esp'
+  - name: '(Unique Region Names|URN.*).esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/11750/' ]
+  - name: 'Unique Region Names.esp'
```
MacSplody, line 6455: Try and keep edits relevant to the pull request.
If this where for updating Unique Region Names or Convenient Carriages then it would be fine.

MacSplody (approved):

masterlist.yaml
sibir-ine: Got it, I'll be more careful about this going forward. Sorry!

PR #1343, opened 2020-10-03
Update Kalilies' NPCs

sibir-ine:
* Added location to patches provided on mod page
  - avoided using catch-all in regex as it would include plugins not on the page
* Added requirement to both CRF patches
* Removed unnecessary clean info I added in #1296 as I now figure it's a bit silly for patches that aren't dirty

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -4226,7 +4226,7 @@ plugins:
         util: 'SSEEdit v4.0.3'
       - crc: 0x062B1FEB
         util: 'SSEEdit v4.0.3'
-  - name: 'Kalilies.*(CRF|Cutting Room Floor) Patch.esp'
+  - name: 'Kalilies.*(CRF|Cutting Room Floor).*\.esp'
     req: [ 'Cutting Room Floor.esp' ]
   - name: 'Kalilies NPCs + AI Overhaul + Cutting Room Floor Patch.esp'
     after: [ 'AI Overhaul - Cutting Room Floor Patch.esp' ]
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -4205,8 +4205,9 @@ plugins:
       - crc: 0xD0F78346
         util: 'SSEEdit v4.0.2f'
 
-  - name: 'KaliliesNPCs.esp'
+  - name: 'Kalilies( )?NPCs( \+ AI Overhaul)?( - CRF| \+ Cutting Room Floor)?( Patch)?.esp'
```
MacSplody: Maybe this instead, a bit easier to look at.
```suggestion
  - name: '(KaliliesNPCs|Kalilies NPCs).*\.esp'
```

sibir-ine: I avoided that since it would include the plugin on [this page](https://www.nexusmods.com/skyrimspecialedition/mods/33942/), `KaliliesNPCs WARP PATCH.esp`. I just figured out how to exclude phrases from a regex, so I'll commit an adjustment for you to take a look at.

MacSplody: Makes sense, but the same thing could happen for any patch hosted on a different page.
Worst case is it's going to have the original mod and patch listed in locations.

sibir-ine: Yeah, that's just the only one I'm aware of and that's linked to on Kalilies' page, so I wanted to take it into consideration. Thankfully, I found a much more elegant solution than my previous one.

MacSplody: It works and isn't hard to read, I'm happy enough with that.

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1344, opened 2020-10-03
Update Blood on the Ice Hotfix

sibir-ine:
Based on the [list](https://www.afkmods.com/index.php?/topic/4682-mods-made-obsolete-by-unofficial-skyrim-special-edition-patch/) of mods made obsolete by USSEP, it apparently can actually break the quest and should be avoided altogether (Arthmoor references the USSEP version explicitly, but the issue he mentions is present in the non-USSEP version as well).

> This file serves absolutely no purpose since the triggers are enabled in aliases during MS11 and will actually BREAK the quest if you use it.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1634,8 +1634,9 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/24941/' ]
     group: *fixesGroup
     msg:
-      - <<: *useInstead
-        subs: [ '[Unofficial Skyrim Special Edition Patch](https://www.nexusmods.com/skyrimspecialedition/mods/266)' ]
+      - <<: *alreadyInOrFixedByX
+        subs: [ 'Unofficial Skyrim Special Edition Patch' ]
+        condition: 'active("Unofficial Skyrim Special Edition Patch.esp")'
 
   - name: 'BugfixCompilation.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/33449/' ]
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1634,9 +1634,8 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/24941/' ]
     group: *fixesGroup
     msg:
-      - <<: *alreadyInOrFixedByX
-        subs: [ 'Unofficial Skyrim Special Edition Patch' ]
-        condition: 'active("Unofficial Skyrim Special Edition Patch.esp")'
+      - <<: *useInstead
+        subs: [ '[Unofficial Skyrim Special Edition Patch](https://www.nexusmods.com/skyrimspecialedition/mods/266)' ]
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -1634,9 +1634,8 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/24941/' ]
     group: *fixesGroup
     msg:
-      - <<: *alreadyInOrFixedByX
-        subs: [ 'Unofficial Skyrim Special Edition Patch' ]
-        condition: 'active("Unofficial Skyrim Special Edition Patch.esp")'
```
MacSplody: I don't really see the need for removing the `alreadyInOrFixedByX`

> Based on the [list](https://www.afkmods.com/index.php?/topic/4682-mods-made-obsolete-by-unofficial-skyrim-special-edition-patch/) of mods made obsolete by USSEP, it apparently can actually break the quest and should be avoided altogether (Arthmoor references the USSEP version explicitly, but the issue he mentions is present in the non-USSEP version as well).
> 
> > This file serves absolutely no purpose since the triggers are enabled in aliases during MS11 and will actually BREAK the quest if you use it.

How does changing the message from

```yaml
  - &alreadyInOrFixedByX
    type: error
    content:
      - lang: en
        text: 'Delete. Already included or otherwise fixed in %1%.'
```

to

```yaml
  - &useInstead
    type: warn
    content:
      - lang: en
        text: 'Use %1% instead.'
```

get across that the file should not be used? And why it shouldn't be used?

masterlist.yaml
```diff
@@ -1634,9 +1634,8 @@ plugins:
...
+      - <<: *useInstead
+        subs: [ '[Unofficial Skyrim Special Edition Patch](https://www.nexusmods.com/skyrimspecialedition/mods/266)' ]
```
MacSplody: I don't mind adding this, but It's unlikely anyone not using the USSEP is going to switch or stop using the file.
A condition needs to be added as well, since it's a warning message.

masterlist.yaml
sibir-ine: Good point, I've kept the alreadyInOrFixedByX message and instead added a useInstead message that includes a link to the forum post discussing why the mod shouldn't be used.

masterlist.yaml
sibir-ine: Fair point, I just figured that since the mod can break the quest, it's worth adding a message even for the handful of people it'd be useful to. I've also added a condition, sorry about that.

MacSplody: Okay, there's no harm in adding it.

masterlist.yaml
MacSplody: > instead added a useInstead message that includes a link to the forum post discussing why the mod shouldn't be used.

It might look a bit weird with two links, but if it gets across the point it's fine.

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1356, opened 2020-10-06
Corrections for Invent.Remove and Relev changes made in #1341

sibir-ine:
* Fixed implementation for removing Invent.Remove from plugins using Invent to accommodate how WB handles deprecated tags
* Added back Relev for mods that change the flags (LVLF) subrecord of leveled lists
  - Will require a manual review when the leveled list patcher is rewritten
* Also noticed that the removal of Invent.Remove doesn't seem to show up next to the plugin in LOOT even though it shows up in its metadata and is recognized by WB. Not sure if it's an issue on my end or a bug in LOOT
plugin
meta

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6724,8 +6724,7 @@ plugins:
       - <<: *patch3rdParty_LotD
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_ArsMetallica_Patch.esp")'
     tag:
-      - -Invent
-      - Invent.Add
+      - -Invent.Remove
       - -Names
     clean:
       - crc: 0x4688FC6A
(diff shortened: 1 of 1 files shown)


MacSplody:
> Also noticed that the removal of Invent.Remove doesn't seem to show up next to the plugin in LOOT even though it shows up in its metadata and is recognized by WB. Not sure if it's an issue on my end or a bug in LOOT

LOOT will only show removal if the tag is in the plugins header.
Technically `Invent.Remove` isn't in the plugins header, only `Invent`.

masterlist.yaml
```diff
@@ -12828,8 +12826,7 @@ plugins:
     inc: [ 'DragonBridgeBoardwalk.esp' ]
     tag:
       - Actors.AIPackages
-      - -Invent
-      - Invent.Add
```
MacSplody, line 12832: You can keep the the `Invent.Add`. It won't make a difference in WB, but it might prompt the author to replace the `Invent` tag.

MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

Issue #1357, opened 2020-10-06
Testing masterlist files with Invent tags


MacSplody:
[Interesting NPCs SE (3DNPC)](https://www.nexusmods.com/skyrimspecialedition/mods/29194/)
- [x] `3DNPC.esp`

[Bibliophile's Arcanaeum](https://www.nexusmods.com/skyrimspecialedition/mods/26772/)
- [x] `BA_(LegacyoftheDragonborn|LostLibrary)\.esp`

[Beasts of Tamriel](https://www.nexusmods.com/skyrimspecialedition/mods/5404/)
- [x] `Beasts of Tamriel.esp`

[Cloaks of Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/6369/)
- [x] `Cloaks.esp`

[Immersive College of Winterhold](https://www.nexusmods.com/skyrimspecialedition/mods/17004/)
- [x] `CollegeOfWinterholdImmersive.esp`, `ICOW_MOW_Patch.esp`.

[Crafting Supplies - All In One](https://www.nexusmods.com/skyrimspecialedition/mods/2359/)
- [x] `CraftingSupplies_AIO_ALL.esp`

[Cutting Room Floor](https://www.nexusmods.com/skyrimspecialedition/mods/276/')
- [x] `Cutting Room Floor.esp`

[Dawnguard Arsenal SSE](https://www.nexusmods.com/skyrimspecialedition/mods/25094/)
- [x] `DawnguardArsenal.esp`

[DLC Integration](https://www.nexusmods.com/skyrimspecialedition/mods/8032/)
- [x] `DLCIntegration.esp`

[Guards Armor Replacer](https://www.nexusmods.com/skyrimspecialedition/mods/22671/)
- [x] `Guards_Armor_Replacer.esp`

[Harvest Overhaul Redone](https://www.nexusmods.com/skyrimspecialedition/mods/2398/)
- [x] `HarvestOverhaul_Redone.esp`, `HarvestOverhaul_Redone_Creatures.esp`.

[The Definitive Hearthfire Supplies Mod](https://www.nexusmods.com/skyrimspecialedition/mods/5993/)
- [x] `Hearthfire Materials.esp`

[Immersive Horses](https://www.nexusmods.com/skyrimspecialedition/mods/13402/)
- [x] `Immersive Horses.esp`

[Immersive Jewelry](https://www.nexusmods.com/skyrimspecialedition/mods/5336/)
[...]

PR #1359, opened 2020-10-06
Update tags for Cutting Room Floor

sibir-ine:
* Removed tags already in plugin description
* Invent is in description, added Invent.Add & Invent.Remove in case author wants to update & replace deprecated tags (#1357)
* Removed unnecessary tags (Actors.DeathItem & C.RecordFlags) in description
* Added EffectStats, EnchantmentStats, Outfits.Add, Outfits.Remove, & SpellStats

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12028,6 +12028,7 @@ plugins:
     tag:
       - -Actors.DeathItem
       - -C.RecordFlags
+      - EnchantmentStats
       - Factions
       - Invent.Add
       - Invent.Remove
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12026,15 +12026,7 @@ plugins:
         condition: 'active("TheChoiceIsYours.esp") and not active("TCIY_CRF_Patch.esp")'
       - *patch3rdPartySRPC
     tag:
-      - C.Location
-      - C.Owner
-      - C.RecordFlags
-      - Delev
       - Factions
-      - Graphics
-      - Invent
-      - Names
-      - Stats
     clean:
       - crc: 0xE71FE395
         util: 'SSEEdit v4.0.3'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12027,6 +12027,8 @@ plugins:
       - *patch3rdPartySRPC
     tag:
       - Factions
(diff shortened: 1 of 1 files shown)


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

Arthmoor:
Referencing the comment in #1542 I already made changes to the tags in the last release:

Actors.ACBS, Actors.AIData, Actors.AIPackages, Actors.CombatStyle, Actors.Spells, Actors.Stats, C.Location, C.Owner, C.Regions, Delev, EffectStats, EnchantmentStats, Factions, Graphics, Invent.Add, Invent.Remove, Keywords, NPC.Class, Names, Outfits.Add, Outfits.Remove, SpellStats, Stats, Text

is the current set, which I had to move to a text file in the Data\BashTags folder because it's now too large for the description field.

MacSplody:
> is the current set, which I had to move to a text file in the Data\BashTags folder because it's now too large for the description field.

Right, we'll have to check those for tags manually as LOOT doesn't know about them.

Ortham:
> Right, we'll have to check those for tags manually as LOOT doesn't know about them.

The bash tags LOOT knows about are sourced from the masterlist's `bash_tags` array.

MacSplody:
> 
> 
> > Right, we'll have to check those for tags manually as LOOT doesn't know about them.
> 
> The bash tags LOOT knows about are sourced from the masterlist's `bash_tags` array.

It compares against the header of the plugins is more what I meant.
So you'll know what tags are being applied overall.
It won't check against these files stored in `Data\BashTags` so you won't see them in LOOT.

Arthmoor:
It might be something to consider writing up support for if more mod authors start adding their tag lists. That description field is limited to 512 chars and a tag list eats that up fast. These external files may end up getting used more.

Ortham:
> It won't check against these files stored in `Data\BashTags` so you won't see them in LOOT.

Ah, I see. In that case it might be worth adding support for `Data\BashTags` then, as Arthmoor says.

Infernio:
See [this comment](https://github.com/loot/loot/issues/1408#issuecomment-770201701), we need to have a discussion about the future of these files.

PR #1382, opened 2020-10-13
Update AI Overhaul - Open Cities patch

sibir-ine:
* Moved to 3rd party patches section as it's not on either mod's page (my bad)
* Removed after as Open Cities is now a master
* Clean as of 0.2
* The entry could also be removed as I originally only added it for the after


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1403, opened 2020-10-16
Update Skyrim Alchemy and Food Overhaul

MacSplody:
#1357

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -7066,6 +7066,8 @@ plugins:
     msg:
       - <<: *corrupt
         condition: 'checksum("SAFO - Another Sorting Mod.esp", EB023317)'
+  - name: 'SAFO - Flora Respawn Fix.esp'
+    req: [ 'FloraRespawnFix.esp' ]
   - name: 'SAFO Leveled Items Fix.esp'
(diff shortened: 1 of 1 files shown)


sibir-ine (commented):
Just a heads up, I tested SAFO with a bashed patch and the leveled list additions get forwarded just fine, even the multiple entries which WB can sometimes have an issue with. I tested both on its own and with a test plugin to overwrite the additions.

MacSplody:
> Just a heads up, I tested SAFO with a bashed patch and the leveled list additions get forwarded just fine, even the multiple entries which WB can sometimes have an issue with. I tested both on its own and with a test plugin to overwrite the additions.

Maybe it's removing empty sublists, I haven't tested with that turned on yet, either way the level list fix wouldn't be required.

MacSplody:
With remove empty sublists turned on.

image

sibir-ine:
Ah, that option completely slipped my mind. Considering it should only be used if the user knows they don't have any mods using leveled list injectors, they shouldn't be using it with this mod. It's disabled by default for that reason, so most people probably won't run into this problem. Also, the fix will completely overwrite any other changes to those 3 leveled lists, so it's probably best avoided.

PR #1416, opened 2020-10-19
Add common alias patchUnavailable

sibir-ine:
* For mods that require a proper patch but don't currently have one.
* Replaced all instances of message with alias.


MacSplody (approved):

MacSplody:
@sibir-ine Thanks.

PR #1496, opened 2020-11-04
Replace Travis with Github Actions

MacSplody:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-16.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get install libstdc++6
+          sudo apt-get install libhttp-parser-dev
+          sudo apt-get install libssh2-1-dev
```
Ortham: These three could be done in one command, which is probably faster.
```
sudo apt-get install -y libstdc++6 libhttp-parser-dev libssh2-1-dev
```

(The `-y` flag should be passed to answer yes to any otherwise-interactive prompts.)

Ortham (changes requested):

.github/workflows/ci.yml
MacSplody: Okay done.

Ortham (approved):

PR #1509, opened 2020-11-08
Update Guard Dialogue Overhaul SE

sibir-ine:
* Update cleaning info
  - Version 2.14.
  - Remove comments as they can be seen in the blame.

### WACCF
* Update after
  - Add checksum for GDO SE 2.14.
  - Remove checksums for older versions of GDO SE to reduce clutter, kept 2.13 and True GDO Port from GDO Redux.


MacSplody:
@sibir-ine Thanks.

PR #1526, opened 2020-11-28
Falkreath Hauntings - Add Do not clean message

TateTaylorOH:
opusGlass' [Falkreath Hauntings](https://www.nexusmods.com/skyrimspecialedition/mods/20733/) contains 3 ITMs. These are required in order to for the Lady Ilinalta encounter to work properly. They are edits to flow data so that the freezing effect that the boss battle happens as is scripted. Water mods and any other mod that edits flow data in that location will cause the water not to freeze and render the boss battle impossible to complete.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5287,10 +5287,7 @@ plugins:
 
   - name: 'ogFalkreathHauntings.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/20733/' ]
-    msg:
-      - type: say
-        content: 'Do not clean. "Dirty" edits are intentional and required for the mod to function.'
-        lang: eng
+    msg: [ *doNotClean ]
 
   - name: 'Rebalanced Encounter Zones.*\.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/25/' ]
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5287,11 +5287,10 @@ plugins:
 
   - name: 'ogFalkreathHauntings.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/20733/' ]
-    dirty:
-      - <<: *quickClean
-        crc: 0xF302A166
-        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
-        itm: 3
+    msg:
+      - type: say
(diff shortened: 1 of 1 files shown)


MacSplody:
Cleaning won't remove the changes to water flow, it does remove two vanilla navmesh though.

image

There is a way to keep those changes from being cleaned by changing the editorID.
But I think It's probably a conflict with another mod, and given the area it's in, that could be a lot of mods.
Moon and Star, Undeath, Tel Nalta, Lake Haven, Millwater Retreat, Oar's Rest to name a few.

image

TateTaylorOH:
Changing the editiorID is a good idea.

Yeah, there's a lot of patches available, could we also get a LOOT warning if those patches aren't detected?

masterlist.yaml
```diff
@@ -5287,11 +5287,10 @@ plugins:
...
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/20733/' ]
-    dirty:
-      - <<: *quickClean
-        crc: 0xF302A166
-        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
-        itm: 3
+    msg:
+      - type: say
+        content: 'Do not clean. "Dirty" edits are intentional and required for the mod to function.'
+        lang: eng
```
MacSplody: You could use the alias doNotClean instead, it does have a slightly different message.

> Do not clean ITM records, they are intentional and required for the mod to function. It is safe to undelete records, but do not do anything other than that.

```suggestion
    msg: [ *doNotClean ]
```

TateTaylorOH: That works for me if it works for you.

MacSplody:
> Changing the editiorID is a good idea.
> 
> Yeah, there's a lot of patches available, could we also get a LOOT warning if those patches aren't detected?

Adding those is pretty straight forward, there are a few different alias types depending on where the patch is hosted.

For example there's the alias `patchProvided` for any patches on the original modpage.
```yaml
  - name: 'Tel_Nalta.esp'
    msg:
      - <<: *patchProvided
        subs: [ 'Moon and Star' ]
        condition: 'active("MoonAndStar_MAS.esp") and not active("MoonAndStar_TelNaltaPatch.esp")'
```

TateTaylorOH:
They should all be hosted on the mod page.

MacSplody:
@PierreDespereaux Thanks.

TateTaylorOH:
Thank _you_!

Issue #927, opened 2019-12-18
LotD v5 Incompatibilities List


MacSplody:
https://www.nexusmods.com/skyrimspecialedition/articles/1494
Some incompatible mods.

_Originally posted by @Xahtax in https://github.com/loot/skyrimse/issues/922#issuecomment-566762999_

---

> This is a running list of mods that have conflicts with Legacy of the Dragonborn V5. They will be removed from the list if and when patches are made, or remain here if they continue to not have support.
> 
> Open Cities Skyrim- Without a patch, the museum does not exist in OCS because it transplants all of the enclosed cities into the open worldspace. We have no plans to patch this because OCS is touchy at best and causes A LOT of insane work to adapt just for the sake of riding a horse in town and saving 1 load door.
> 
> Creation Club- The Legacy team members do not own any CC content and therefore we cannot patch it. If you wish to patch it, contact SirJesto on the Legacy Patch Central page and we may pack your patch in our FOMOD. There already is content for most CC items in the community, most of which look or function better, and many of which are already in Legacy.
> 
> Dawnguard/Clab Volkihar Epilogues- Legacy already has a solution built into it to provide all of the path specific relics from the dawnguard questlines. This mod is not needed and may cause conflicts
> 
> Solitude Skyway- Due to Legacy's overhaul of the arch, Solitude Skyway will require a remodel patch to work around it. There is not way to patch Legacy to work with it.
> 
> The People of Skyrim- An included Tree LOD sticks up in front of the museum and cannot be patched out. You can fix this by re-running DynDOLOD or SSELODGen

Xahtax:
MAS Immersion Patch is not compatible with LotD

Xahtax:
- [x] Open Cities

- [x] Dawnguard/Clan Volkihar Epilogues

- [x] Solitude Skyway

- [x] MAS Immersion patch

Won't add the Creation Club and The People of Skyrim

MacSplody:
> Won't add the Creation Club and The People of Skyrim

That's fine, a link to compatibility list could be added to LotD to cover those.

```yaml
  - name: 'LegacyoftheDragonborn.esm'
    msg:
      - <<: *compatNotes
        subs: [ 'https://www.nexusmods.com/skyrimspecialedition/articles/1494' ]
```

Xahtax:
I added it.

MacSplody:
[...]

PR #1532, opened 2020-12-18
Removing false information: Dawnguard Epilogues

SkyLover264:
As the mod author of Dawnguard and Clan Volkihar Epilogues (Dawnguard&VolkiharArtifactsQuests.esp), I can tell you with certainty that Legacy of the Dragonborn, and my mod, are not incompatible. I've used them side-by-side since I made it, even with V5 of Legacy, and they work together without a problem.

Likewise, the 4 itms are false reports, and are needed for the mod to work.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9718,7 +9718,11 @@ plugins:
       # version:  1.0r
       - crc: 0x53109251
         util: 'SSEEdit v4.0.2f'
-
+        
+  - name: 'Dawnguard&VolkiharArtifactsQuests.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12098' ]
+    msg: [ *doNotClean ]
+    
   - name: 'Dwarfsphere.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/15996/' ]
     msg:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9718,7 +9718,6 @@ plugins:
       # version:  1.0r
       - crc: 0x53109251
         util: 'SSEEdit v4.0.2f'
-        
   - name: 'Dawnguard&VolkiharArtifactsQuests.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12098' ]
     msg: [ *doNotClean ]
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9719,15 +9719,6 @@ plugins:
       - crc: 0x53109251
         util: 'SSEEdit v4.0.2f'
 
-  - name: 'Dawnguard&VolkiharArtifactsQuests.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12098' ]
-    inc: [ *LotD_v5 ]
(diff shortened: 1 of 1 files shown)


pStyl3:
The incompatible message was issued in this issue: https://github.com/loot/skyrimse/issues/927
The information comes from this [article](https://www.nexusmods.com/skyrimspecialedition/articles/1494) from the Legacy of the Dragonborn SSE Nexus page.
```
Dawnguard/Clan Volkihar Epilogues - Legacy already has a solution built into it to provide all of the path specific relics from the Dawnguard questlines. Do not install this mod.
```

SkyLover264:
I originally responded to the Legacy team's assessment on my own mod page. I'll repost it here:

"Over the past year, I've seen some misinformation spreading that this mod is not compatible with Legacy of the Dragonborn. I would like to shoot that down now. Aside from some duplicate items, this mod and Legacy do NOT interfere with each other in any way, shape, or form. In fact, my mod provides some things that Legacy does not. I've decided to address this here as I am sick and tired of correcting every post I see claiming this. Henceforth, I ask that you not continue to spread this misinformation any further."

themayor897 (changes requested):
It's irresponsible to completely remove the entry from the masterlist rather than edit it. Also, I think it's wise to wait for the LOTD team to weigh in on this before committing any changes.

masterlist.yaml
```diff
@@ -9719,15 +9719,6 @@ plugins:
       - crc: 0x53109251
         util: 'SSEEdit v4.0.2f'
 
-  - name: 'Dawnguard&VolkiharArtifactsQuests.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/12098' ]
-    inc: [ *LotD_v5 ]
```
themayor897, line 9724: msg: [ *doNotClean ]

masterlist.yaml
```diff
@@ -9719,15 +9719,6 @@ plugins:
...
-    dirty:
```
themayor897, line 9725: if the donotclean message is added lines 9725-end should be removed.

themayor897 (changes requested):

MacSplody:
There's a compromise which can be reached regarding the views on compatibility.

I added a message in https://github.com/loot/skyrimse/commit/e464a908dbd10f409b6fb651374dccbff3bcbd35 which could work here. It would need to be added to the `LegacyoftheDragonborn.esm` metadata.

SkyLover264:
> It's irresponsible to completely remove the entry from the masterlist rather than edit it. Also, I think it's wise to wait for the LOTD team to weigh in on this before committing any changes.

I don't think their word would be trustworthy here. They hate me.

masterlist.yaml
```diff
@@ -9718,16 +9718,11 @@ plugins:
       # version:  1.0r
       - crc: 0x53109251
         util: 'SSEEdit v4.0.2f'
-
+
```
MacSplody: No tabs should be used, only spaces.
```suggestion

```

MacSplody (approved):

MacSplody:
@SkyLover264 Thanks.

PR #1533, opened 2020-12-24
Wheels of Lull 5.0.0 Update

themayor897:
Removed patches that are no longer needed and not included in the installer.
Updated crc to reflect the newest version of the mod.
Added bash tag to add a relationship to a faction that is edited by multiple mods (Thalmor)


MacSplody (approved):

MacSplody:
@themayor897 Thanks.

PR #1534, opened 2020-12-26
Sort Apocalypse after Ordinator if patched.

Gavvers:
The Apocalypse compatibility patch for Ordinator expects Apocalypse to be loaded after Ordinator. This sorting is only applied if the patch is actually used.


MacSplody (approved):

MacSplody:
@Gavvers Thanks.

PR #1552, opened 2021-01-29
Add warning for Immersive Jewelry

### LINKED ISSUE #1536: Adding a warning to Immersive Jewelry?
Right now it breaks the BP, because it wasn't converted properly to SSE and so its WEAP\CRDT subrecords have a really weird size (20 bytes, whereas 16 is the LE version and 24 is the SE version).
Resaving it in the CK seems to fix this, since it doesn't actually use any critical effects that resaving can break (they're all NULL).

Infernio:
Version 1.05 contains very funky WEAP\CRDT subrecords. They're in SE format, but the last 4 bytes are truncated. Version 1.06a on the mod page fixes this.

The main problem with these records is that they make WB choke, but they could cause problems ingame as well.

Closes #1536.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -11184,6 +11184,13 @@ plugins:
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_ImmersiveJewelry_Patch.esp")'
       - <<: *patch3rdParty_KPH_CCOR
         condition: 'active("Complete Crafting Overhaul_Remastered.esp") and not active("Immersive Jewelry_CCOR_Patch.esp")'
+      - type: warn
+        content:
+          - lang: en
+            text: 'Version 1.05 of this mod contains corrupt records. Update to version 1.06a.'
+          - lang: de
+            text: 'Die Version 1.05 dieses Mods enthält korrupte Einträge. Aktualieren Sie zu Version 1.06a.'
+        condition: 'version("Immersive Jewelry.esp", "1.05", ==)'
     tag:
       - Delev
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -11184,6 +11184,13 @@ plugins:
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_ImmersiveJewelry_Patch.esp")'
       - <<: *patch3rdParty_KPH_CCOR
         condition: 'active("Complete Crafting Overhaul_Remastered.esp") and not active("Immersive Jewelry_CCOR_Patch.esp")'
+      - type: warn
+        content:
+          - lang: en
+            text: 'Version 1.05 of this mod contains corrupt records. Update to version 1.06a.'
+          - lang: de
+            text: 'Die Version 1.05 dieses Mods enthält korrupte Einträge. Aktualieren Sie zu Version 1.06a.'
+        condition: 'version("Immersive Jewelry.esp", "1.05", ==)'
```
MacSplody, line 11193: Could use a combination of *Corrupt and *Obsolete, but if you think this message is better then I'm fine with it.

MacSplody (approved):

masterlist.yaml
Infernio: Considered it, but I think being more specific with the version is good here because 1.06a is only in the Updates sections on the Nexus. The main file is still 1.05, aka the broken one.

MacSplody:
@Infernio Thanks.

PR #1560, opened 2021-02-01
Added Ultimate College of Winterhold

themayor897:
Added data to Immersive College of Winterhold and Obscure's College of Winterhold to reflect that there is now a patch for them.

Added section for Ultimate College of Winterhold since I will be adding a bunch of data for patches as I make them.


MacSplody (approved):
@themayor897 That's fine, is that all you want to add?

themayor897:
@MacSplody as of now, yes, it gets the main conflicts out of the way and looks good when I tested it on my client. I will do another pull request for the ICOW clean status and patch. Thank you so much for all your help!

MacSplody:
@themayor897 Thanks.

PR #1580, opened 2021-02-08
Swedish Localization

ddbb07:
I think i translated everything but if there's something i missed, tell me.


MacSplody:
> 
> 
> I think i translated everything but if there's something i missed, tell me.

That's all of the common and global messages covered for Swedish.

There are other messages, but what you've done can be added as it is.

The other messages are generally one offs and don't qualify for common at the moment.

image

MacSplody (approved):

MacSplody:
@ddbb07 Thanks.

PR #1584, opened 2021-02-09
Add "Text" bashed tag to AlchemyAdjustments.esp

ubeogesh:
The tag is needed to properly display potion duration for potions that are instant in vanilla (e.g. "restore 6 health per second for 5 seconds" - the "for 5 seconds" part was lost, as it was imported from unofficial patch instead)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -7281,6 +7281,7 @@ plugins:
       - 'EconomyOverhaulandSpeechcraftImprovementsLite.esp'
   - name: 'AlchemyAdjustments.esp'
     after: [ 'SkyTEST-RealisticAnimals&Predators.esp' ]
+    tag: [ Text ]
 
   - name: 'Ars Metallica.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -7281,6 +7281,8 @@ plugins:
       - 'EconomyOverhaulandSpeechcraftImprovementsLite.esp'
   - name: 'AlchemyAdjustments.esp'
     after: [ 'SkyTEST-RealisticAnimals&Predators.esp' ]
+    tag:
+      - Text
```
sibir-ine: Since it's just the one tag, I'd recommend inline formatting as follows:
```suggestion
    tag: [ Text ]
```

ubeogesh: done

MacSplody (approved):

MacSplody:
@ubeogesh Thanks.

PR #1592, opened 2021-02-12
Update to Serana Dialogue Addon

ddbb07:
- Added Patch Messages

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9364,11 +9364,29 @@ plugins:
         subs: [ '[Serana Dialogue Edit](https://www.nexusmods.com/skyrimspecialedition/mods/16222/)' ]
         condition: 'not active("Serana Dialogue Edit.esp")'
       - <<: *patchProvided
+        subs: [ 'Amorous Adventures' ]
+        condition: 'active("SexLab-AmorousAdventures.esp") and not active("SDA-AA Patch.esp")'
+      - <<: *patchProvided
         subs: [ 'DX Crimson Blood Armor' ]
         condition: 'active("Crimson Blood Armor.esp") and not active("SDA- DX CBlood Patch.esp")'
       - <<: *patchProvided
         subs: [ 'Lustmord Vampire Armor' ]
         condition: 'active("LustmordVampireArmor.esp") and not active("SDA-LustmordPatch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Nether''s Follower Framework' ]
+        condition: 'active("nwsFollowerFramework.esp") and not active("SDA NFF Patch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Relationship Dialogue Overhaul' ]
+        condition: 'active("Relationship Dialogue Overhaul.esp") and not active("SDA RDO Patch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Serana Dialogue Edit' ]
+        condition: 'active("Serana Dialogue Edit.esp") and not active("SDA_SDE Patch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Wintersun - Faiths of Skyrim' ]
+        condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Complete Alchemy & Cooking Overhaul.esp") and(not active("SDA Wintersun Patch.esp") or not checksum("SDA Wintersun Patch.esp", A0EB4F89))'
+      - <<: *patchProvided
+        subs: [ 'Wintersun and CACO' ]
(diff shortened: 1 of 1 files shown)


sibir-ine:
Some of the patches are purely optional so it might be best to not include messages for them. I added a few a while back that probably also count for this, so, if @MacSplody thinks it's appropriate, do you mind removing them as well & only keeping the ones that are necessary? I'm thinking patches like the ones for the armor, but some others might fall under this umbrella too. Also, could you remove the alsoUseX message while your at it? I originally added it, but I don't really think the author's recommendation warrants the message anymore. No problem if you don't want to, I can in another PR.

MacSplody:
> Some of the patches are purely optional so it might be best to not include messages for them. I added a few a while back that probably also count for this, so, if @MacSplody thinks it's appropriate, do you mind removing them as well & only keeping the ones that are necessary? I'm thinking patches like the ones for the armor, but some others might fall under this umbrella too.

I'm leaning more towards leaving optional patches out, but there's no real harm in adding them, so I have no objections.
Up to you @ddbb07 on whether you agree with @sibir-ine's point or not.

masterlist.yaml
```diff
@@ -9363,12 +9363,30 @@ plugins:
...
+        condition: 'active("Relationship Dialogue Overhaul.esp") and not active("SDA RDO Patch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Serana Dialogue Edit' ]
+        condition: 'active("Serana Dialogue Edit.esp") and not active("SDA_SDE Patch.esp")'
+      - <<: *patchProvided
+        subs: [ 'Wintersun - Faiths of Skyrim' ]
+        condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Complete Alchemy & Cooking Overhaul.esp") and(not active("SDA Wintersun Patch.esp") or not checksum("SDA Wintersun Patch.esp", A0EB4F89))'
+      - <<: *patchProvided
+        subs: [ 'Wintersun and CACO' ]
+        condition: 'active("Wintersun - Faiths of Skyrim.esp") and active("Complete Alchemy & Cooking Overhaul.esp") and(not active("SDA Wintersun Patch.esp") or not checksum("SDA Wintersun Patch.esp", 2F889D8A))'
```
MacSplody: My suggestion from Discord.

```suggestion
      - <<: *patchProvided
        subs: [ 'Wintersun - Faiths of Skyrim' ]
        condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Complete Alchemy & Cooking Overhaul.esp") and not active("SDA Wintersun Patch.esp")'
      - <<: *patchProvided
        subs: [ 'Wintersun and CACO' ]
        condition: 'active("Wintersun - Faiths of Skyrim.esp") and active("Complete Alchemy & Cooking Overhaul.esp") and not active("SDA Wintersun Patch.esp")'
```

With a separate message for the wrong patch being used.

```yaml
  - name: 'SDA Wintersun Patch.esp'
    msg:
      - <<: *useVersion
        subs: [ 'Complete Alchemy & Cooking Overhaul' ]
        condition: 'active("Complete Alchemy & Cooking Overhaul.esp") and checksum("SDA Wintersun Patch.esp", A0EB4F89)'
```
If you prefer your way that's fine, I just think it's a bit clearer this way.

ddbb07: Im sorry for not fixing this earlier i have just not had alot of time, i'll mark it as a draft until im done

MacSplody: > Im sorry for not fixing this earlier i have just not had alot of time, i'll mark it as a draft until im done

There's generally no rush when submitting changes, so don't worry about it.

ddbb07:
I changed optional patches from patchProvided to alsoUseX, if you don't like the change just leave a comment.

masterlist.yaml
```diff
@@ -9361,19 +9361,40 @@ plugins:
...
     tag: [ Keywords ]
     clean:
       - crc: 0x61AFBCBD
         util: 'SSEEdit v4.0.3'
-
+    
+  - name: 'SDA Wintersun Patch.esp'
+    msg:
+      - <<: *useVersion
+        subs: [ 'Wintersun and CACAO Patch' ]
```
sibir-ine: This fits the message a bit better:
```suggestion
        subs: [ 'Complete Alchemy & Cooking Overhaul' ]
```
Either that or:
```suggestion
        subs: [ 'CACO' ]
```
Whichever you prefer.

sibir-ine (changes requested):
Other than my suggestion, my only concern is that if someone has several of these mods, it may feel a bit cluttered to have multiple messages for optional patches they can't dismiss without installing them. However, this may be addressed in a future version of LOOT as described in loot/loot#1083, so this is fine for now.

sibir-ine (approved):

sibir-ine:
Thanks!

PR #1600, opened 2021-02-18
Beyond Skyrim: Bruma & Identity Crisis

megapatato:
[Identity Crisis](https://www.nexusmods.com/skyrimspecialedition/mods/39634)
[Beyond Skyrim: Bruma](https://www.nexusmods.com/skyrimspecialedition/mods/10917)

With both mods loaded, the game is largely broken, with players trapped within interiors. This is apparently [known behavior](https://forums.nexusmods.com/index.php?/topic/5818982-beyond-skyrim-bruma-se/page-396#entry80205298) (4th item in the Common Issues list).

The note _"a compatibility patch is provided at X"_ is technically accurate, so I used that, but is there a _"yell at players to install this or their game will be very broken"_ giant popup box of blinking red text..?

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -12260,6 +12260,10 @@ plugins:
         condition: 'active("Complete Alchemy & Cooking Overhaul.esp") and active("ccbgssse037-curios.esl") and not active("CACO_Bruma_Rare Curios_Patch.esp")'
       - <<: *patch3rdPartyUVCPC
         condition: 'not active("Vokrii - Bruma Patch.esp") and active("Vokrii - Minimalistic Perks of Skyrim.esp")'
+      - <<: *patchProvided
+        subs:
+          - 'Identity Crisis'
+        condition: 'active("BSHeartland.esm") and active("Identity Crisis.esp") and not active("BS Bruma Patch.esp")'
     tag:
       - Actors.ACBS
       - Actors.CombatStyle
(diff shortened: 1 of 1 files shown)


sibir-ine:
Since the patch is hosted on Identity Crisis' page, the `patchIncluded` message should be put in an entry for `Identity Crisis.esp` & updated accordingly. Since one doesn't exist, you'll have to create it. You can append it to the end of the masterlist if your unsure where it should go.

After that, you can remove one of the `active` checks in the condition statement, namely the one for `Identity Crisis.esp`, as it's redundant though harmless.

Finally, if you think the incompatibility is dire enough, you can add an `inc` entry with a condition to hide it when the patch is active. Here's an example for adding it to an entry for `Identity Crisis.esp`:
```yaml
  - name: 'Identity Crisis.esp'
    inc:
      - name: 'BSHeartland.esm'
        display: 'Beyond Skyrim - Bruma SE'
        condition: 'not active("BS Bruma Patch.esp")'
```

megapatato:
I see. That is reasonable, and to avoid injecting unnecessary commits I'll fork/pull-request anew with your suggestion.

sibir-ine:
> 
> 
> I see. That is reasonable, and to avoid injecting unnecessary commits I'll fork/pull-request anew with your suggestion.

For future reference, you don't need to worry about extra commits in the PR. It will be squashed and committed as one commit anyway. Whatever works best for you.

PR #1619, opened 2021-02-26
Update Castle Volkihar Rebuilt

ddbb07:
- Added patch messages to "Castle Volkihar Rebuilt.esp"
- Updated clean message for "Castle Volkihar Rebuilt.esp"
- Added use instead message to "SFCO - Castle Volkihar Rebuilt Patch.esp"


sibir-ine:
Thanks!

PR #1625, opened 2021-02-27
Add Perks Unbound

megapatato:
[Perks Unbound](https://www.nexusmods.com/skyrimspecialedition/mods/1800) changes perk unlocking conditions.

[WACCF](https://www.nexusmods.com/skyrimspecialedition/mods/18994) changes... a whole lotta things, including keyword conditions for material effects or friendly fire checks.

One overwrites the other. [The patch](https://www.nexusmods.com/skyrimspecialedition/mods/46295) bases from WACCF and applies Unbound's condition removals.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -18845,12 +18845,12 @@ plugins:
         condition: 'active("CCOR_Vokrii.esp") and active("Vokrii - WACCF Patch.esp")'
 
 ###### Anything new can go after this (If you're not sure where to put it) ######
-# Perks Unbound -- Weapons Armor Clothing & Clutter Fixes patch
+ # Perks Unbound -- Weapons Armor Clothing & Clutter Fixes patch
   - name: 'Perks Unbound - Special Edition.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/1800' ]
     msg:
       - <<: *patch3rdParty
         subs: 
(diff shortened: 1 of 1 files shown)


sibir-ine:
@megapatato You could add the message to the `Weapons Armor Clothing & Clutter Fixes.esp` entry instead so you don't have to make a new entry for `Perks Unbound - Special Edition.esp`. It'd look like this:
```yaml
      - <<: *patch3rdParty
        subs:
          - 'Perks Unbound'
          - '[Perks Unbound SSE -- WACCF patch](https://www.nexusmods.com/skyrimspecialedition/mods/46295/)'
        condition: 'active("Perks Unbound - Special Edition.esp") and not active("PerksUnbound_WACCF.esp")'
```
Whichever you prefer.

megapatato:
I have no real preference either way. I guess I'd slightly prefer not changing it because changing a thing is technically more work than doing nothing ¯\_(ツ)_/¯

For future reference, all else being equal, is it preferable to append to an entry already existing, or to create a new entry?

sibir-ine:
> For future reference, all else being equal, is it preferable to append to an entry already existing, or to create a new entry?

It doesn't really matter either way, it's just a bit tidier to add a patch message to an entry that already has patch messages.

sibir-ine:
Thanks!

PR #1627, opened 2021-02-27
Update Armor & Clothing Extension

such-dev:
Fixes for Armor and Clothing Extension now seems included in 1.1 of Armor and Clothing Extension and therefore not required anymore.


sibir-ine:
Thanks!

PR #1643, opened 2021-03-02
Update Messages & globals urls to use https

pStyl3:


masterlist.yaml
```diff
@@ -1003,7 +1003,7 @@ globals:
   - <<: *versionXIncY
     subs:
       - '**SKSE Plugin ([ENB Helper SE](https://www.nexusmods.com/skyrimspecialedition/mods/23174/))**'
-      - '[ENBSeries](http://enbdev.com/download_mod_tesskyrimse.html)'
+      - '[ENBSeries](https://enbdev.com/download_mod_tesskyrimse.html)'
```
sibir-ine, line 1006: The ENB site doesn't seem to support HTTPS. Also, that link is outdated. It's `http://enbdev.com/download_mod_tesskyrimse.htm` now.

masterlist.yaml
```diff
@@ -1014,7 +1014,7 @@ globals:
   # ENBSeries
   - <<: *missingRequirementXforY
     subs:
-      - '[ENBSeries](http://enbdev.com/download_mod_tesskyrimse.html)'
+      - '[ENBSeries](https://enbdev.com/download_mod_tesskyrimse.html)'
```
sibir-ine, line 1017: Same issue here.

masterlist.yaml
pStyl3: I have no problem opening that website using `https`, I checked it before changing it.
Besides that, the url can be updated of course.

sibir-ine: It works, it's just technically not correct.

pStyl3: I guess you're right. Will change it in another PR.

PR #1650, opened 2021-03-05
Add race tags to Imperious

Infernio:
Added in the latest WB nightly. Finally managed to refactor this beast.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -7163,6 +7163,10 @@ plugins:
         condition: 'not active("Universal Imperious Patch.esp")'
     tag:
       - Names
+      - R.ChangeSpells
+      - R.Description
+      - R.Skills
+      - R.Stats
       - Text
     msg:
       - <<: *compatNotes


sibir-ine:
Thanks for all your guys' hard work on the patcher! I can already think of at least one other plugin off the top of my head that these'll be useful for.

MacSplody:
These tags might need to be disabled when using [Universal Imperious Patch SE](https://www.nexusmods.com/skyrimspecialedition/mods/38109). I haven't done any testing of it yet, but that patch basically makes it so the race changes from Imperious aren't required.

Infernio:
Had a quick look, most of its changes can be handled by tags. E.g. It changes some height (`R.Body-Size-F`/`R.Body-Size-M`), alters Imperious' race descriptions (`R.Description`) and carries forward the skills and stats changes from Imperious (no tags needed since Imperious has them and this mod requires Imperious as a master). The only thing that might be problematic are the spells. It undoes the spell changes that Imperious does, so I'm not sure if `R.ChangeSpells` on the patch would work, you'd have to test it.

PR #1651, opened 2021-03-07
Update Arcanum - A New Age of Magic

themayor897:
Added information about BSA that needs to be renamed or the mod will be broken. Added link to patch that fixes various Oldrim assets included in this version.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8590,14 +8590,15 @@ plugins:
   - name: 'Arcanum.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/23488/' ]
     msg:
-      - <<: *patch3rdPartySRPC
+      - *patch3rdPartySRPC
       - <<: *alsoUseX
-        subs: '[Arcanum - A New Age of Magic 4.0.1 Bugfix](https://www.nexusmods.com/skyrimspecialedition/mods/46322)'
-        condition: 'checksum("Arcanum.esp", AA67A039)
+        subs: [ '[Arcanum - A New Age of Magic 4.0.1 Bugfix](https://www.nexusmods.com/skyrimspecialedition/mods/46322)' ]
+        condition: 'checksum("Arcanum.esp", AA67A039)'
       - type: warn
-        content: 'The BSA files included with this version of Arcanum need to be renamed in order for the mod to work. Remove "4.0 BSA" from the name of each .bsa file so you are left with Arcanum.bsa and Arcanum - Textures.bsa.'
-        lang: eng
-        condition: 'checksum("Arcanum.esp", AA67A039)
+        content: 
+          - lang: en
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -8589,12 +8589,21 @@ plugins:
...
-    msg: [ *patch3rdPartySRPC ]
+    msg:
+      - *patch3rdPartySRPC
+      - <<: *alsoUseX
+        subs: [ '[Arcanum - A New Age of Magic 4.0.1 Bugfix](https://www.nexusmods.com/skyrimspecialedition/mods/46322)' ]
+        condition: 'checksum("Arcanum.esp", AA67A039)'
+      - type: warn
+        content: 
+          - lang: en
+            text: 'The BSA files included with this version of Arcanum need to be renamed in order for the mod to work. Remove "4.0 BSA" from the name of each .bsa file so you are left with Arcanum.bsa and Arcanum - Textures.bsa.'
```
MacSplody: I added something similar to this to the Oblivion masterlist.
So it could be used instead as it has some translations.

```yaml

  - &renameXtoY
    type: warn
    content:
      - lang: en
        text: 'You need to rename %1% to %2% or this mod won''t work.'
      - lang: de
        text: 'Sie müssen %1% zu %2% umbenennen, ansonsten funktioniert diese Mod nicht.'
      - lang: ja
        text: '%1%を%2%にリネームしないと機能しません。'
      - lang: ru
        text: 'Вам нужно переименовать %1% в %2% для подключения ресурсов мода..'
```
```yaml
  - name: 'MigHorseArmor.esp'
    msg:
      - <<: *renameXtoY
        subs:
          - 'HRMHorseArmor.bsa'
          - 'MigHorseArmor.bsa'
        condition: 'not file("MigHorseArmor.bsa") and file("HRMHorseArmor.bsa")'
```

You'd have to use it twice though.

masterlist.yaml
```diff
@@ -8589,12 +8589,21 @@ plugins:
...
+        condition: 'checksum("Arcanum.esp", AA67A039)'
```
```yaml
        condition: 'checksum("Arcanum.esp", AA67A039)'
```
The issue with using `Arcanum.esp` in the condition is it will always show the message unless they modify that file.
It should really be checking if the wrong BSA file is installed, and also make sure a valid version of the BSA isn't installed.
```yaml
        condition: 'not file("Arcanum.bsa") and file("Arcanum 4.0 BSA.bsa")'
```

themayor897: Oh nice! I thought you could only use plugin names, not bsas. Thank you!

sibir-ine: > 
> 
> Oh nice! I thought you could only use plugin names, not bsas. Thank you!

You can read more about what conditions you can use [here](https://loot-api.readthedocs.io/en/stable/metadata/conditions.html).

sibir-ine (approved):

sibir-ine:
Thanks!

PR #1652, opened 2021-03-08
Update to UCOW

ddbb07:
- Added patch message for Apophysis

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -18814,7 +18814,7 @@ plugins:
         condition: 'active("NewArmoury.esp") and not active("UltimateCollege_NewArmoury_Patch.esp")'
       - <<: *patchIncluded
         subs: [ 'Apophysis Dragon Priest Masks' ]
-        condition: '(active("Apophysis_DPM_SE.esp") or file("textures\carn2k4\dragon priest masks\hevnoraak_hood.dds")) and not file("meshes/CWI/DisplayStatics/CWIHevnoraak.nif")'
+        condition: '(active("Apophysis_DPM_SE.esp") or file("textures/carn2k4/dragon priest masks/hevnoraak_hood.dds")) and not file("meshes/CWI/DisplayStatics/CWIHevnoraak.nif")'
       - <<: *patchIncluded
         subs: [ 'Arcanum' ]
         condition: 'active("Arcanum.esp") and not active("UltimateCollege_Arcanum_Patch.esp")'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -18812,6 +18812,9 @@ plugins:
       - <<: *patchIncluded
         subs: [ 'Animated Armoury' ]
         condition: 'active("NewArmoury.esp") and not active("UltimateCollege_NewArmoury_Patch.esp")'
+      - <<: *patchIncluded
+        subs: [ 'Apophysis Dragon Priest Masks' ]
+        condition: 'active("Apophysis_DPM_SE.esp") and not checksum("meshes\CWI\DisplayStatics\CWIHevnoraak.nif", 8B653612)'
```
```suggestion
        condition: '(active("Apophysis_DPM_SE.esp") or file("textures/carn2k4/dragon priest masks/hevnoraak_hood.dds")) and not file("meshes/CWI/DisplayStatics/CWIHevnoraak.nif")'
```
@ddbb07 I'd add a check for one of the textures from Apophysis since there's a loose version of the mod without the plugin.

Also, it should be enough to check if one of the patch files is just present since the unpatched version will be in Immersive CoW's BSA.

Lastly, you need to use forward slashes for paths.

sibir-ine (changes requested):

masterlist.yaml
ddbb07: I don't think forward slashes are necessary, i tested it with normal slahses and it worked fine but i'll change the condition.

sibir-ine: @ddbb07 They have to be forward slashes; I tested to double-check, & back slashes cause a syntax error.

Also sorry, I accidentally used back slashes for the other path & didn't realize.

ddbb07: Your right but the first version didn't require forward slashes so i don't know why that worked.

sibir-ine: No clue how it passed the check, but you can always [test your own branch](https://loot.github.io/docs/contributing/Quickly-Testing-Your-Masterlist-Changes.html) to know for sure.

ddbb07: It didn't just pass the test it also did what i wanted it to, and i know how to test my branch

sibir-ine: Ah, my bad. It's because `checksum` functions just take a `path` argument while `file` functions take a `regex`. That means you can do either forward or backslashes for `checksum` functions but you have to use forward slashes for `file` functions as back slashes are special characters. That's why it passed the first time but failed all the other checks.

It'd be best to use forward slashes for `checksum` functions anyway for the sake of consistency. Nearly every other instance of the `checksum` function uses forward slashes, & it'd also be consistent with `file` functions.

sibir-ine (approved):

sibir-ine:
Thanks! Sorry for the confusion.

PR #1653, opened 2021-03-10
Update Wheels of Lull

themayor897:
Removed Wheels of Lull's Bash tags since they're already included in the plugin description. Updated to latest CRC. Added a writing patch someone made for Wheels of Lull to the masterlist that is now incompatible because it edits old records, and we have meticulously fixed typos so it is probably obsolete as well. Author has not responded to me.

I will add a version number to Lull's description when we next update it for more stable conditions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+Skyrim Special Edition - Shortcut.lnk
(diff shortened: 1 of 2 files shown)


sibir-ine (approved):
Yeah, version checks are much better than checksum checks. They'll do for now though. You can read more about how LOOT handles plugin versioning [here](https://loot-api.readthedocs.io/en/stable/metadata/conditions.html).

sibir-ine:
Thanks!

PR #1677, opened 2021-03-17
Add Cathedral - 3D Pine Grass load order

sean-kang:
The author of 3D Pine Grass says the mod should be loaded after grass mods like Cathedral Landscape and Cathedral 3D mountain flower.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -5973,6 +5973,11 @@ plugins:
       - crc: 0xA4E835CA
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Cathedral - 3D Pine Grass.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/42032/' ]
+    after:
+      - 'Cathedral Landscapes.esp'
+
   - name: 'cliffs.esp'
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/4520'


masterlist.yaml
```diff
@@ -5973,6 +5973,12 @@ plugins:
       - crc: 0xA4E835CA
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Cathedral - 3D Pine Grass.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/42032/' ]
+    after:
+      - 'Cathedral Landscapes.esp'
+      - 'Cathedral - 3D Mountain Flowers.esp'
```
sibir-ine: `Cathedral - 3D Mountain Flowers.esp` is a master, so `Cathedral - 3D Pine Grass.esp` will always load after. Just in case, is there another version of `Cathedral - 3D Mountain Flowers.esp` other than the one hosted [here](https://www.nexusmods.com/skyrimspecialedition/mods/41312/) that isn't a master?

masterlist.yaml
```diff
@@ -5973,6 +5973,12 @@ plugins:
       - crc: 0xA4E835CA
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Cathedral - 3D Pine Grass.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/42032/' ]
+    after:
+      - 'Cathedral Landscapes.esp'
```
sibir-ine, line 5979: LOOT sorts `Cathedral - 3D Pine Grass.esp` after `Cathedral Landscapes.esp` by default, but it wouldn't hurt to add a rule anyway if you still want to.

sibir-ine (dismissed):

masterlist.yaml
sean-kang: Right, it is esm-flagged. I removed it in the update.

masterlist.yaml
sean-kang: Hmm. It seems to work that way. If there is an implicit rule that sorts `Cathedral - 3D Pine Grass.esp` after `Cathedral Landscapes.esp`, I will cancel the pull request.

MacSplody (approved):

masterlist.yaml
MacSplody: If you look in the file header for `Cathedral - 3D Pine Grass.esp`, you'll see its masters.
Plugins will always load after their master files.

Like below.
image

sean-kang: There is no dependency between those two mods, that is defined in the header. I was asking if there was a hidden rule that somehow sorted the load order correctly.

MacSplody: > There is no dependency between those two mods, that is defined in the header. I was asking if there was a hidden rule that somehow sorted the load order correctly.

No there's not a hidden rule, but plugins are ordered using a sorting algorithm.
The sorting algorithm determined it was the optimal order.

MacSplody:
@sean-kang Thanks.

masterlist.yaml
sibir-ine: @sean-kang Sorry for the confusion! When I said it sorted after by default, I was referring to the sorting algorithm Mac mentioned.

Issue #646, opened 2019-03-13
Translations for messages


MacSplody:
If the language you want to translate for isn't listed, see [this article](https://github.com/loot/loot/blob/master/CONTRIBUTING.md#translating-loot) to add support for your language to LOOT.
If it is, feel free to check if the localization for your language is up-to-date anyway.

While pull requests are appreciated, you can simply leave a comment with your translations.
Any contribution is always appreciated.

**NOTE**
_Actual translations only, don't just use google translate._

---

**Overall progress**
  - [x] Bulgarian (bg) 🏆
  - [ ] Czech (cs)
  - [ ] Danish (da)
  - [x] German (de) 🏆
  - [ ] Spanish (es)
  - [ ] Finnish (fi)
  - [ ] French (fr)
  - [x] Japanese (ja) 🏆
  - [ ] Korean (ko)
  - [ ] Polish (pl)
  - [ ] Portuguese (pt)
  - [x] Portuguese Brazilian (pt_BR) 🏆
  - [ ] Russian (ru)
  - [ ] Swedish (sv)
  - [ ] Simplified Chinese (zh_CN)

---

### _common_

**&alreadyInX**

`'Delete. Already included in %1%.'`

**&alreadyInOrFixedByX**

`'Delete. Already included or otherwise fixed in %1%.'`

**&alsoUseX**

`'It''s highly recommended that you also use %1%.'`

**&compatIssuesWithX**

`This plugin has compatibility issues with %1%. For more information, read this mods compatibility notes.`

**&compatNotes**

`'It is recommended that you read this mod''s [Compatibility Notes](%1%).'`

**&corrupt**

`'This file is corrupt and should not be used.'`

**&deletePlugin**

`'When using **%1%**, it''s recommended that you deactivate or delete this ESP file but keep the resources (e.g. meshes, textures) installed with this mod.'`

**&includesX**

`'%1% is included with this mod.'`

**&missingRequirementXforY**

`'It appears you have installed **%2%**, but some of its requirements seem to be missing. Please ensure you have correctly installed **%1%**.'`

**&missingRequirementXforPlugin**

`'Some of this plugins requirements seem to be missing. Please ensure you have correctly installed **%1%**.'`

**&obsolete**

`'Obsolete. Update to the latest version. %1%'`

**&optional**

`'This plugin is optional.'`

**&patch3rdParty**

`'You seem to be using **%1%**, but you have not enabled a compatibility patch for this mod. A third party patch is available here: %2%'`

[...]

PR #1682, opened 2021-03-18
Add Mysticism patch message

sean-kang:
It adds a message for Mysticism - FleshFX patch and also removes the outdated load order for "MM - CRF + USSEP.esp".


sibir-ine:
@sean-kang Thanks! If you don't mind, could you take a look at [this comment](https://github.com/loot/skyrimse/issues/646#issuecomment-801038795)?

PR #1688, opened 2021-03-22
Fix broken link for Falskaar - Unique Region Names

jordanbtucker:
When `Falskaar - Unique Region Names.esp` is loaded, LOOT gives the warning to use `Unique Region Names - Falskaar addon` instead, but the provided link is broken because it is seen as a relative URL. This commit makes the URL absolute.


sibir-ine:
Thanks! Can't believe I missed this when I merged the changes a while ago 🙃

PR #1709, opened 2021-03-29
Add Water for ENB

MacSplody:


masterlist.yaml
```diff
@@ -3728,6 +3728,25 @@ plugins:
...
+  - name: 'Water for ENB( \((Mineral Teal|Tropical Green|Wavy Rivers)\))?\.esp'
+    inc:
+      - 'iNeed.esp'
+      - 'Keep It Clean.esp'
+      - 'RealisticNeedsandDiseases.esp'
+      - 'tdgsseWadeInWater.esp'
+    msg:
+      - <<: *useVersion
+        subs: [ 'Needs' ]
+        condition: 'active("(iNeed|RealisticNeedsandDiseases|Keep It Clean|tdgsseWadeInWater)\.esp")'
```
sibir-ine, line 3748: In WB 308, @Infernio added a tweak, `Tweak Assorted -> Mark All Water As Damaging`, that's designed to do what most water mod - needs mod patches do. Haven't checked if it applies in this case, but we should look into other patches & add messages accordingly. This case would be a bit more complicated since it's a replacer instead of a patch, but we could probably do something like hide the message & `inc` entries if a BP is present & recommend they also use the tweak.

MacSplody: > In WB 308, @Infernio added a tweak, `Tweak Assorted -> Mark All Water As Damaging`, that's designed to do what most water mod - needs mod patches do. Haven't checked if it applies in this case, but we should look into other patches & add messages accordingly. This case would be a bit more complicated since it's a replacer instead of a patch, but we could probably do something like hiding the message & `inc` entries if a BP is present & recommend they also use the tweak.

It's going to be tough to do that reliably, and account for the order they install the mods.
We also have no way to tell what version of WB is being used, or if a tweak has been used.
We could switch the inc list for a `compatIssuesWithX` message so that they won't get an error.
Adding the **also use tweak message** in the needs mods metadata might be fine.

PR #1720, opened 2021-03-30
Update to Wyrmstooth

ddbb07:
- Removed old versions
- Updated CRC's
- Updated version numbers
- Added forward slashes to the end of links (just because of personal preference)
- Removed dirty message

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -13295,7 +13295,7 @@ plugins:
       - <<: *obsolete
         subs: [ '[Wyrmstooth](https://www.nexusmods.com/skyrimspecialedition/mods/45565/)' ]
         condition: 'not version("Wyrmstooth.esp", "1.19.2", >=)'
-    clean: # CRC's for Drachenzahn (Wyrmstooth) DV 1.19.2, Wyrmstooth 1.19.2
+    clean:
       - crc: 0x7D46DF7E
         util: 'SSEEdit v4.0.3'
       - crc: 0x4FB18D3A
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -13281,12 +13281,10 @@ plugins:
 
   - name: 'Wyrmstooth.esp'
     url:
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/45565'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/45565/'
         name: 'Wyrmstooth'
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/19492'
-        name: 'Drachenzahn (Wyrmstooth) EV 2.2'
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/18553'
-        name: 'Drachenzahn (Wyrmstooth) DV 3.0'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -13296,20 +13294,15 @@ plugins:
...
         subs: [ '[Wyrmstooth](https://www.nexusmods.com/skyrimspecialedition/mods/45565/)' ]
-        condition: 'not version("Wyrmstooth.esp", "1.19", >=)'
-    dirty:
-      - <<: *quickClean
-        crc: 0xF73A6161
-        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
-        itm: 5587
-    clean:
+        condition: 'not version("Wyrmstooth.esp", "1.19.2", >=)'
+    clean: # CRC's for 1: Unknown, 2: Unknown, 3: Drachenzahn (Wyrmstooth) DV 1.19.2, 4: Wyrmstooth 1.19.2 #
```
MacSplody: We can put those in the commit message like below.
image


> CRC's for 1: Unknown, 2: Unknown, 3: Drachenzahn (Wyrmstooth) DV 1.19.2, 4: Wyrmstooth 1.19.2 #

```suggestion
    clean:
```

masterlist.yaml
```diff
@@ -13296,20 +13294,15 @@ plugins:
...
       - crc: 0x97F8C8B4
         util: 'SSEEdit v4.0.3'
       - crc: 0x85D4BF27
         util: 'SSEEdit v4.0.3'
```
MacSplody: You can remove the cleaning info for these old versions.

masterlist.yaml
```diff
@@ -13281,12 +13281,10 @@ plugins:
...
     url:
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/45565'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/45565/'
         name: 'Wyrmstooth'
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/19492'
-        name: 'Drachenzahn (Wyrmstooth) EV 2.2'
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/18553'
-        name: 'Drachenzahn (Wyrmstooth) DV 3.0'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/18553/'
+        name: 'Drachenzahn (Wyrmstooth) DV 1.19.2'
```
```suggestion
        name: 'Drachenzahn (Wyrmstooth) DV'
```

MacSplody (changes requested):

masterlist.yaml
ddbb07: I'd prefer keeping something like that so when people update the crc's they know which one to update

MacSplody: > I'd prefer keeping something like that so when people update the crc's they know which one to update

Yeah, I used to do it that way, but the commit messages can be viewed in git blame or with git lens if you're using VS code.
It's reliable and they can't be removed as easily as comments.

image

image

ddbb07: I'll remove it then

MacSplody (approved):

MacSplody:
@ddbb07 Thanks.

PR #1732, opened 2021-04-03
Update BOTI.esp metadata

BrainBlasted:
Blood on the Ice Redux is _not_ incompatible with Finding Susana Alive. In fact, the author recommends that they are paired together:

> ...If you want to prevent this problem for good, check SomethingObscure's mod Finding Susanna Alive.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -10348,7 +10348,6 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6126/'
         name: 'Blood on the Ice Redux'
     after: [ 'Quest Conflict Fixes.esp' ]
-    inc: [ 'FSA_FindingSusannaAlive.esp' ]
     msg:
       - <<: *alsoUseX
         subs: [ '[Fuz Ro D-oh - Silent Voice](https://www.nexusmods.com/skyrimspecialedition/mods/15109)' ]


masterlist.yaml
```diff
@@ -10348,7 +10348,6 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6126/'
         name: 'Blood on the Ice Redux'
     after: [ 'Quest Conflict Fixes.esp' ]
-    inc: [ 'FSA_FindingSusannaAlive.esp' ]
```
sibir-ine, line 10351: This version of FSA is incompatible with BotIR, so the `inc` entry should be kept.

sibir-ine (changes requested):
@BrainBlasted The vanilla version of Finding Susanna Alive is incompatible with Blood on the Ice Redux. However, there is a BotIR-compatible version available in FSA's installer. ~~You can add an entry for FSA with a message to warn BotIR users to use that version of FSA like so:~~

```yaml
  - name: 'FSA_FindingSusannaAlive.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/32512/' ]
    msg:
      - <<: *useVersion
        subs: [ 'Blood on the Ice Redux' ]
        condition: 'active("BOTI.esp")'
```

~~You can add it to the `Gameplay - Quests` section or the end of the masterlist.~~

I'll add FSA in a separate PR & let you know when it's merged.

BrainBlasted:
My bad; I don't know how I missed that before.

sibir-ine:
> 
> 
> My bad; I don't know how I missed that before.

@BrainBlasted No worries, thank you for pointing the issue out!

PR #1746, opened 2021-04-07
Added Open Cities - LOTD Compatibility Patch

ddbb07:
- Added "OCS + LOTD5 Patch.esp"


sibir-ine:
@ddbb07 We prefer not to add cleaning data for patches or small add-ons if they're clean since it's excessive, so I'll close this. If there's other relevant metadata for this patch, feel free to open a new PR to submit it.

PR #1748, opened 2021-04-08
Update to patch3rdParty messages

ddbb07:
- Matched naming on the Nexus with naming in the masterlist
- Added links for the source mod's
- Added forward slashes to the end of links

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -400,43 +400,43 @@ common:
   - &patch3rdParty_BAOPFCH
     <<: *patch3rdParty
     subs:
-      - 'Convenient Horses'
+      - '[Convenient Horses](https://www.nexusmods.com/skyrimspecialedition/mods/9519/)'
       - '[Bruma and Other Patches for Convenient Horses](https://www.nexusmods.com/skyrimspecialedition/mods/13812/)'
   - &patch3rdParty_CCSMPC
     <<: *patch3rdParty
     subs:
       - 'CC Survival Mode'
-      - '[Survival Mode Patch Collection](https://www.nexusmods.com/skyrimspecialedition/mods/19154)'
+      - '[Survival Mode Patch Collection](https://www.nexusmods.com/skyrimspecialedition/mods/19154/)'
   - &patch3rdParty_JSDCP
     <<: *patch3rdParty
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -450,24 +450,24 @@ common:
...
     <<: *patch3rdParty
     subs:
       - '[Scoped Bows](https://www.nexusmods.com/skyrimspecialedition/mods/912/)'
-      - '[Scoped Bows Compatibility Kit - SBCK](https://www.nexusmods.com/skyrimspecialedition/mods/19495)'
+      - '[Scoped Bows Compatibility Kit - SBCK](https://www.nexusmods.com/skyrimspecialedition/mods/19495/)'
   - &patch3rdPartySRPC
     <<: *patch3rdParty
     subs:
-      - '[Spell Research v2.2+](https://www.nexusmods.com/skyrimspecialedition/mods/20983/)'
+      - '[Spell Research](https://www.nexusmods.com/skyrimspecialedition/mods/20983/)'
```
sibir-ine: The patches on that page are specifically for version 2.2 & up of that mod, so we should specify that in the sub.
```suggestion
      - '[Spell Research v2.2+](https://www.nexusmods.com/skyrimspecialedition/mods/20983/)'
```

sibir-ine (changes requested):
@ddbb07 The links to the mods aren't really necessary since the user will already have them installed & probably know where they got them from, but we can add them anyway.

ddbb07:
> @ddbb07 The links to the mods aren't really necessary since the user will already have them installed & probably know where they got them from, but we can add them anyway.

Why i added the links to the mods was mostly for standardisation since some messages had them and others didn't but i also think it's better to have them than not.

sibir-ine:
> Why i added the links to the mods was mostly for standardisation since some messages had them and others didn't but i also think it's better to have them than not.

@ddbb07 Yeah, it's fine. The only issue is with the changes to `&patch3rdPartySRPC`. Revert those & it'll be good to go.

sibir-ine (approved):

sibir-ine:
@ddbb07 Thanks!

PR #1768, opened 2021-04-12
Update to various clean messages

ddbb07:
- Updated clean message for SeranaDialogAddon.esp
- Updated patch crc for SeranaDialogAddon.esp
- Updated clean message for Ordinator
- Updated clean message for Project Proteus
- Updated clean message for Undeath Classic Lichdom


MacSplody:
@ddbb07 Thanks.

If you're updating the same thing for different mods, then consider splitting changes for each mod into separate commits.
You can still do it all in one pull request, but having it like https://github.com/loot/skyrimse/pull/649/commits would be best.

image

PR #1918, opened 2021-07-07
Remove Birds of Skyrim

MacSplody:
Mod deleted.
https://www.nexusmods.com/skyrimspecialedition/mods/3097/

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6744,17 +6744,6 @@ plugins:
       - Relev
 
 ###### Encounters - Neutrals ######
-  - name: 'BirdsOfSkyrim_SSE.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/3097/' ]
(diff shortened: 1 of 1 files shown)


ghost:
Not sure what LOOT policy about non Nexus mods is, but all of steve40's mods are available on his Dropbox. He also mentioned that he might move them to a new platform "soon". Links and info on his profile: https://www.nexusmods.com/users/3463396

pStyl3:
@ColonSpeedruns 
If steve40 uploads his mods to a new hosting service, we certainly can reevaluate adding the meta data regarding his mods once again. I don't think we should start adding Dropbox links to our masterlists, as there is little external control for such files. This doesn't mean that we don't trust steve40, it just means that for LOOT it would be better to link to files that are uploaded to a dedicated hosting service, where we can see what the mod is all about and what else we should consider while maintaining the meta data for those mods.

PR #1955, opened 2021-07-24
Update Wheels of Lull

themayor897:
Added incompatibility with old add-on since it always causes problems and we've retired it.

Added cleaning info for new version.


masterlist.yaml
```diff
@@ -12358,9 +12358,14 @@ plugins:
         condition: 'active("DA_Skyship.esm") and not active("WheelsofLull-Skyship.esp")'
       - <<: *patch3rdParty_LotD
         condition: 'active("LegacyoftheDragonborn.esm") and not active("DBM_WheelsofLull_Patch.esp") and (version("LegacyoftheDragonborn.esm", "5.0.24", >=) or (not checksum("LegacyoftheDragonborn.esm", 31563183) and not checksum("LegacyoftheDragonborn.esm", 6736FB94) and not checksum("LegacyoftheDragonborn.esm", CB82480D)))'
+    inc:
+      - name: 'WoLStartAt40.esp'
+        condition: version("WheelsOfLull.esp", "5.1.0", >=)
```
```suggestion
        condition: 'version("WheelsOfLull.esp", "5.1.0", >=)'
```

MacSplody (changes requested):
Just need to wrap that condition and it's fine then.

MacSplody (approved):

MacSplody:
@themayor897 Thanks.

PR #1958, opened 2021-07-26
Wheels of Lull high poly head patch

themayor897:
Old versions of this patch will cause an infinite loading screen when used with new versions of Lull, so I thought it important to add this info.


masterlist.yaml
```diff
@@ -12366,6 +12369,13 @@ plugins:
         util: 'SSEEdit v4.0.3'
       - crc: 0xAC83F041
         util: 'SSEEdit v4.0.3'
+
```
MacSplody: I'll just remove that newline, plugins from the same mod page can be kept together.
```suggestion
```

MacSplody (approved):

MacSplody:
@themayor897 Thanks.

PR #1963, opened 2021-08-03
Require new plugin-provided patch for TCIY

fwdekker:
[The Choice is Yours](https://www.nexusmods.com/skyrimspecialedition/mods/3850) (TCIY) and [Book Covers Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/901/) (BCS) require a patch to work together. Previously, this patch was found on [kryptopyr's Patch Hub](https://www.nexusmods.com/skyrimspecialedition/mods/19518/), but the patch is now provided by TCIY.

This PR
1. changes the name of the patch from `TCIY BCS Patch.esp` to `TCIY_BCS Patch.esp`,
2. changes the message from `patch3rdParty_KPH_TCIY` to `patchIncluded`, and
3. moves the patch from BCS to TCIY.

Additionally, this PR still allows a bashed patch to resolve the conflict instead (not tested, but based on old patch specification). I also see that there's a patch specification for `Legacyofthedragonborn.esm` right above the new patch specification that has the eerily similar name `BCS_TCIY_Patch.esp`, but I did not touch that because I did not use Legacy of the Dragonborn.

I have verified locally that the message is displayed correctly when the patch is absent or the old patch is used, and that the message is not displayed if the new patch is used.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -13180,7 +13180,7 @@ plugins:
           - '[Wintersun - More Patches](https://www.nexusmods.com/skyrimspecialedition/mods/24228/)'
         condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Wintersun - BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
       - <<: *patch3rdParty_KPH_TCIY
-        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
+        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
     after:
       - 'Tel_Nalta.esp'
       - 'UnlimitedBookshelves.esp'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -13180,7 +13180,7 @@ plugins:
           - '[Wintersun - More Patches](https://www.nexusmods.com/skyrimspecialedition/mods/24228/)'
         condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Wintersun - BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
       - <<: *patch3rdParty_KPH_TCIY
-        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
+        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch\.esp") and not active("Bashed Patch.*\.esp")'
     after:
       - 'Tel_Nalta.esp'
       - 'UnlimitedBookshelves.esp'


MacSplody:
> [The Choice is Yours](https://www.nexusmods.com/skyrimspecialedition/mods/3850) (TCIY) and [Book Covers Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/901/) (BCS) require a patch to work together. Previously, this patch was found on [kryptopyr's Patch Hub](https://www.nexusmods.com/skyrimspecialedition/mods/19518/), but the patch is now provided by TCIY.

The patch is on both pages, it has a slightly different name on each mod page, but both are the same in content.

masterlist.yaml
```diff
@@ -12308,6 +12308,9 @@ plugins:
     msg:
       - <<: *patch3rdParty_LotD
         condition: 'active("LegacyoftheDragonborn.esm") and not active("BCS_TCIY_Patch.esp")'
+      - <<: *patchIncluded
+        subs: [ 'Book Covers Skyrim' ]
+        condition: 'active("Book Covers Skyrim.esp") and not active("TCIY_BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
```
```suggestion
```

masterlist.yaml
```diff
@@ -13179,8 +13182,6 @@ plugins:
           - 'Wintersun - Faiths of Skyrim'
           - '[Wintersun - More Patches](https://www.nexusmods.com/skyrimspecialedition/mods/24228/)'
         condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Wintersun - BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
-      - <<: *patch3rdParty_KPH_TCIY
-        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
     after:
```
MacSplody, line 13184:
```suggestion
      - <<: *patch3rdParty_KPH_TCIY
        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
    after:
```

MacSplody (changes requested):
It would be best to still use the KPH link for the patch message, since that's more likely to be updated. But we can add a check for both file names as they have the same content.

fwdekker:
You're right, I hadn't noticed that KPH still had a non-outdated patch. I've implemented your suggestions 👍

masterlist.yaml
```diff
@@ -13180,7 +13180,7 @@ plugins:
           - '[Wintersun - More Patches](https://www.nexusmods.com/skyrimspecialedition/mods/24228/)'
         condition: 'active("Wintersun - Faiths of Skyrim.esp") and not active("Wintersun - BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
       - <<: *patch3rdParty_KPH_TCIY
-        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
+        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch.esp") and not active("Bashed Patch.*\.esp")'
```
sibir-ine: @FWDekker Since the expression is a regex now, the period should be escaped:

```suggestion
        condition: 'active("TheChoiceIsYours.esp") and not active("TCIY( |_)BCS Patch\.esp") and not active("Bashed Patch.*\.esp")'
```

MacSplody (approved):

PR #1964, opened 2021-08-07
Updated URL data for my mods.

Arthmoor:
Hopefully I've done this right. :P

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -3406,7 +3406,7 @@ plugins:
       - 'SoS - Civilization.esp'
       - 'SoS - The Dungeons.esp'
       - 'SoS - The Wilds.esp'
-	  - 'SoundsofSkyrimComplete.esp'
+      - 'SoundsofSkyrimComplete.esp'
     tag:
       - C.Acoustic
     clean:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -14078,7 +14085,10 @@ plugins:
         util: 'SSEEdit v4.0.3'
 
   - name: 'Skyshards.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/6331/' ]
+    req: [ 'Unofficial Skyrim Special Edition Patch.esp' ]
+    url:
+      - 'https://www.afkmods.com/index.php?/files/file/1939-eso-skyshards/'
+	  - 'https://bethesda.net/en/mods/skyrim/mod-detail/3326552/'
```
pStyl3: This url needs to be indented the same length than the url above, so two empty spaces compared to `url` are needed.

pStyl3:
Thanks!

sibir-ine:
@Arthmoor Just FYI, if a mod has another mod as a master, you don't need to add a requirement message for it; LOOT will display a missing masters message automatically. I also noticed that **Sounds of Skyrim - Complete Edition.esp** already has **C.Acoustic** in its description.

Arthmoor:
Ah, ok. I didn't know the information wasn't needed so figured I'd add it while I was at it.

PR #1960, opened 2021-07-28
Define load order for WICO and Heights of Skyrim

fwdekker:
I found that loading `WICO - Immersive People.esp` before `Heights_of_Skyrim.esp` resulted in some NPCs having black faces. Changing that order resolved the issue, so I created an entry for the two mods in the master list.

To verify that this is indeed necessary, I ran the following procedure:
1. Install [WICO](https://www.nexusmods.com/skyrimspecialedition/mods/2136) (VNL 0.9f) (core files only during installer)
2. Install [Heights of Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/52057) (v1.02) (ESP or ESP-FE does not matter)
3. Use the following load order:
    ```
    <Skyrim + DLC>
    WICO - Immersive Character.esp
    WICO - Immersive People.esp
    Heights_of_Skyrim.esp
    ```
4. Start a new game.
5. Execute `player.moveto 0001a68d` in the console as soon as possible, during the first cutscene.
6. Once something is visible, do the moveto again and immediately execute `tfc`.
7. Move around a bit to find Andurs. **He incorrectly has a black face.**
8. Exit the game and change the load order to
    ```
    <Skyrim + DLC>
    WICO - Immersive Character.esp
    Heights_of_Skyrim.esp
    WICO - Immersive People.esp
    ```
9. Repeat steps 4 through 7. **Andurs no longer has a black face.**

To verify that my patch is correct, I repeated steps 1 through 9 above, but replaced step 8 by using LOOT to sort the entries using the masterlist in this PR.

PR #1965, opened 2021-08-09
 Define load order for WICO and HoS patch

fwdekker:
Similar to #1960: WICO should load after not only HoS, but also the newly-introduced HoS-USSEP patch.


MacSplody:
@FWDekker Thanks.

PR #2092, opened 2021-10-01
MCoW fake-esm conflict with CollegeEntry fix

rafiwui:
Only force CollegeEntry.esp to load after MCoW if MCoW is not tagged as master.
If MCoW is a master, load the additional esp from MCoW (MCoW_EndofLoadOrderOverwrite.esp) after CollegeEntry

Check this thread for reference on the problem:
[MCoW problem](https://forums.nexusmods.com/index.php?/topic/4994715-simple-college-of-winterhold/page-201#entry97501048)


MacSplody (approved):

MacSplody:
@rafiwui Thanks.

Issue #2128, opened 2021-10-19
ITM report, 5 mods


IllusiveMan196:
- [x] [Undriel_QRR_Clockwork_Patch](https://www.nexusmods.com/skyrimspecialedition/mods/20851) (record 04038A48 in 040142EC cell sub-block 0, 1 0500EED7)
```
  - name: 'Undriel_QRR_Clockwork_Patch.esp'
    dirty:
      - <<: *reqManualFix
        crc: 0xF3661E6A
        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        nav: 1
```
- [x] [Good-folks of Raven Rock](https://www.nexusmods.com/skyrimspecialedition/mods/27987)
```
  - name: 'GFoRR.esp'
    dirty:
      - <<: *quickClean
        crc: 0x52E6D27B
        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        udr: 13
    clean:
      - crc: 0x2455CEDA
        util: 'SSEEdit v4.0.3'
```
- [x] [Isle of Artaeum Season 3](https://www.nexusmods.com/skyrimspecialedition/mods/33150)
```
  - name: 'Artaeum.esp'
    dirty:
      - <<: *reqManualFix
        crc: 0x78537FA2
        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        nav: 2
```
- [x] [SDA SDE Patch](https://www.nexusmods.com/skyrimspecialedition/mods/32161)
```
  - name: 'SDA SDE Patch.esp'
    dirty:
      - <<: *quickClean
        crc: 0xE74E3C33
        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        itm: 5
    clean:
      - crc: 0x1B07ECC4
        util: 'SSEEdit v4.0.3'
```
- [x] Bathing Beauties or Beefcake - Luxury Suite SE SKSE64 Patch (was [here](https://www.nexusmods.com/skyrimspecialedition/mods/5646) before taken off of Nexus by migal, now resides [here](https://migal3d.com/bbls/))
```
  - name: 'BBLS_SKSE64_Patch.esp'
    dirty:
      - <<: *quickClean
        crc: 0x03D81D4F
[...]

PR #2156, opened 2021-11-04
Add Undriel Raven Rock Clockwork patch

IllusiveMan196:
Add mod
Add location
Add cleaning data
#2128

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -19568,6 +19568,13 @@ plugins:
       # version: 1.6
       - crc: 0xEAEAF38B
         util: 'SSEEdit v4.0.2f'
+  - name: 'Undriel_QRR_Clockwork_Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/20851/' ]
+    dirty:
+      - <<: *reqManualFix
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -19568,6 +19568,13 @@ plugins:
       # version: 1.6
       - crc: 0xEAEAF38B
         util: 'SSEEdit v4.0.2f'
+  - name: 'Undriel_QRR_Clockwork_Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/20851/' ]
+    dirty:
+      - <<: *reqManualFix
+        crc: 0xF3661E6A
```
sibir-ine: I think you might've edited your copy of this plugin at some point, the checksum I get from a fresh install is different.

```suggestion
        crc: 0x8A98BC76
```

sibir-ine (changes requested):

masterlist.yaml
IllusiveMan196: Yeah, it is marked as ESL in my load list...

sibir-ine (approved):

PR #2159, opened 2021-11-04
Add BBLS mod line

IllusiveMan196:
Add mods
Add location
Add cleaning data
#2128

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -20048,6 +20048,34 @@ plugins:
         util: '[SSEEdit v4.0.3g](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 32
 
+  - name: 'BBLuxurySuite.esm'
+    url: [ 'https://migal3d.com/bbls/' ]
+    dirty:
+      - <<: *quickClean
+        crc: 0x3E29B84D
+        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 5
+  - name: 'BBLuxurySuiteENM.esp'
+    url: [ 'https://migal3d.com/bbls/' ]
+    clean:
+      - crc: 0x679B347E
+        util: 'SSEEdit v4.0.3'
+  - name: 'BBLS_SKSE64_Patch.esp'
+    url: [ 'https://migal3d.com/bbls/' ]
+    dirty:
+      - <<: *quickClean
+        crc: 0x03D81D4F
+        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 132
+        udr: 17
+  - name: ''BBLSmusic.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -20071,6 +20071,34 @@ plugins:
...
+        util: 'SSEEdit v4.0.3'
+  - name: 'BBLS_SKSE64_Patch.esp'
+    url: [ 'https://migal3d.com/bbls/' ]
+    dirty:
+      - <<: *quickClean
+        crc: 0x03D81D4F
+        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 132
+        udr: 17
+  - name: ''BBLSmusic.esp'
```
```suggestion
  - name: 'BBLSmusic.esp'
```

masterlist.yaml
```diff
@@ -20071,6 +20071,34 @@ plugins:
...
+      - crc: 0x679B347E
+        util: 'SSEEdit v4.0.3'
+  - name: 'BBLS_SKSE64_Patch.esp'
+    url: [ 'https://migal3d.com/bbls/' ]
+    dirty:
+      - <<: *quickClean
+        crc: 0x03D81D4F
+        util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 132
+        udr: 17
```
sibir-ine: I also get a different checksum (& ITM count) with a fresh download.

sibir-ine (changes requested):

masterlist.yaml
IllusiveMan196: Which one is it? All of them??? If to believe Migal's website these files are latest versions (of course! Latest versions... I have previous one because that was broken in some places. Add both?)

IllusiveMan196: Can we add both BBLS versions to list? New one was kinda broken for a lot of players so... I haven't updated.

sibir-ine: > Which one is it? All of them???

Just **BBLS_SKSE64_Patch.esp**. You can check what my comment was referring to in the **Files changed** tab on GitHub.

> Can we add both BBLS versions to list? New one was kinda broken for a lot of players so... I haven't updated.

If it's broken enough, the old one can be added. However, is it still available for download? If not, then it shouldn't be added.

IllusiveMan196: If I remember correctly compatibility with some... "immersion" mods was broken, sometimes to the point of not working at all and possibly some core functions of mod as well (it was in Nexus comments, which now maybe are left in a form of forum messages)

sibir-ine: Well, since older versions don't appear to be publicly available anymore, you'll have to either add the cleaning data for the latest version of **BBLS_SKSE64_Patch.esp** or omit it if you don't think it'll be useful.

IllusiveMan196: Updated it

sibir-ine (approved):

sibir-ine:
Thanks!

PR #2190, opened 2021-11-16
Update Beyond Skyrim - Bruma

IllusiveMan196:
Add loadorder as per [sticky comment on Nexus page](https://www.nexusmods.com/skyrimspecialedition/mods/10917?tab=posts) (let it sort after both files, BSAssets.esm supposed to be global (?) assetpack for project)


sibir-ine:
*BSHeartland.esm* has *BSAssets.esm* as a master & *BS_DLC_patch.esp* has *BSAssets.esm* & *BSHeartland.esm* as masters, so LOOT will sort them this way by default.

PR #2191, opened 2021-11-17
Update Additional Clockwork

BellCubeDev:
All of this in under ~~an~~ two hours? Let's hope it holds together after review by people much more experienced than myself.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15945,6 +15945,7 @@ plugins:
       # version: 1.4
       - <<: *quickClean
         crc: 0x616730B8
+        util: '[SSEEdit v4.0.0](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 26
 
   - name: 'DwemerWorldGate.esp'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15873,7 +15873,7 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     clean:
       - crc: 0x6ED1EE37
-        util: 'SSEEdit v4.0.3'
+        util: 'SSEEdit v4.0.4'
   - name: 'ClockworkCourierCounterspell.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     clean:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15891,7 +15891,7 @@ plugins:
         condition: 'active("ClockworkBugFixes.esp")'
     clean:
       - crc: 0xCAE2B6AE
-        util: 'SSEEdit v4.0.3'
+        util: 'SSEEdit v4.0.4'
   - name: 'ClockworkShadowmarksCastleExt.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     msg:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15885,6 +15885,10 @@ plugins:
         util: 'SSEEdit v4.0.3'
   - name: 'ClockworkMausoleumDummyDoorFix.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
+    msg:
+      - <<: *alreadyInX
+        subs: [ '"Additional Clockwork - Bug Fixes" (ClockworkBugFixes.esp)' ]
+        condition: 'active("ClockworkBugFixes.esp")'
     clean:
       - crc: 0xCAE2B6AE
         util: 'SSEEdit v4.0.3'
@@ -22284,6 +22288,10 @@ plugins:
   # Clockwork Sorting Patches #
   - name: 'ClockworkSorting_Food_CACO.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
+    msg:
+      - <<: *alreadyInOrFixedByX
+        subs: [ '"Additional Clockwork - Superior Sorting" (ClockworkSuperiorSorting.esp)' ]
+        condition: 'active("ClockworkSuperiorSorting.esp")'
     clean:
       - crc: 0x83840648
         util: 'SSEEdit v4.0.3'
@@ -22291,6 +22299,9 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     inc: [ 'ClockworkSorting_Ing_CCCurios.esp' ]
     msg:
+      - <<: *alreadyInOrFixedByX
+        subs: [ '"Additional Clockwork - Superior Sorting" (ClockworkSuperiorSorting.esp)' ]
+        condition: 'active("ClockworkSuperiorSorting.esp")'
       - <<: *alreadyInX
         subs: [ '"Additional Clockwork - CACO & Rare Curios Ingredient Sorting Patch" (ClockworkSorting_Ing_CCCurios_CACO.esp)' ]
         condition: 'active("ClockworkSorting_Ing_CCCurios_CACO.esp")'
@@ -22305,6 +22316,9 @@ plugins:
   - name: 'ClockworkSorting_Ing_CCCurios.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     msg:
+      - <<: *alreadyInOrFixedByX
+        subs: [ '"Additional Clockwork - Superior Sorting" (ClockworkSuperiorSorting.esp)' ]
+        condition: 'active("ClockworkSuperiorSorting.esp")'
       - <<: *alreadyInX
         subs: [ '"Additional Clockwork - CACO & Rare Curios Ingredient Sorting Patch" (ClockworkSorting_Ing_CCCurios_CACO.esp)' ]
         condition: 'active("ClockworkSorting_Ing_CCCurios_CACO.esp")'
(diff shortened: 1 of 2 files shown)


masterlist.yaml
```diff
@@ -15869,11 +15869,29 @@ plugins:
         itm: 122
 
   # Additional Clockwork
+  - name: 'ClockworkBugFixes.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
+    clean:
+      - crc: 0x6ED1EE37
+        util: 'SSEEdit v4.0.3'
```
sibir-ine: xEdit v4.0.4 was recently released, so that should be used for new cleaning reports.

```suggestion
    clean:
      - crc: 0x6ED1EE37
        util: 'SSEEdit v4.0.4'
```

masterlist.yaml
```diff
@@ -15869,11 +15869,29 @@ plugins:
...
   - name: 'ClockworkCourierCounterspell.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     clean:
       - crc: 0x6ABC5154
         util: 'SSEEdit v4.0.3'
+  - name: 'ClockworkInterestingInhabitants.esp'
+    clean:
+      - crc: 0x9E77EAE9
+        util: 'SSEEdit v4.0.3'
```
sibir-ine: I get a different checksum, is this a future version of the plugin?

```suggestion
    clean:
      - crc: 0xC73837D9
        util: 'SSEEdit v4.0.4'
```

masterlist.yaml
```diff
@@ -15869,11 +15869,29 @@ plugins:
...
+  - name: 'ClockworkMausoleumDummyDoorFix.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
+    msg:
+      - <<: *alreadyInX
+        subs: [ '"Additional Clockwork - Bug Fixes" (ClockworkBugFixes.esp)' ]
+        condition: 'active("ClockworkBugFixes.esp")'
+    clean:
+      - crc: 0xCAE2B6AE
+        util: 'SSEEdit v4.0.3'
```
```suggestion
    clean:
      - crc: 0xCAE2B6AE
        util: 'SSEEdit v4.0.4'
```

masterlist.yaml
```diff
@@ -15906,16 +15924,18 @@ plugins:
     clean:
       - crc: 0xB4C1E773
         util: 'SSEEdit v4.0.3'
-  - name: 'ClockworkTitleRemover.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
+  - name: 'ClockworkSuperiorSorting.esp'
+    req: [ *PapyrusExtender, *SKSE ]
```
sibir-ine: To keep formatting consistent across masterlists.

```suggestion
    req:
      - *PapyrusExtender
      - *SKSE
```

masterlist.yaml
```diff
@@ -15906,16 +15924,18 @@ plugins:
...
+    msg:
+      - <<: *scriptExtenderRequiredScripts
```
sibir-ine: This message is already displayed as a global message, so it doesn't need to be displayed for specific plugins.

```suggestion
```

masterlist.yaml
```diff
@@ -15906,16 +15924,18 @@ plugins:
...
     clean:
-      - crc: 0x913191E9
+      - crc: 0x80123223
         util: 'SSEEdit v4.0.3'
```
sibir-ine: I get a different checksum, is this a future version of the plugin?

```suggestion
      - crc: 0x9E78E8D9
        util: 'SSEEdit v4.0.4'
```

masterlist.yaml
```diff
@@ -15906,16 +15924,18 @@ plugins:
...
-  - name: 'ClockworkMausoleumDummyDoorFix.esp'
+  - name: 'ClockworkTitleRemover.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
     clean:
-      - crc: 0xCAE2B6AE
-        util: 'SSEEdit v4.0.3g'
+      - crc: 0xBF1FBE1F
+        util: 'Human Review in SSEEdit v4.0.3' # It doesn't like my deleted records, but I can veryfiy that they don't cause any issues in-game.
```
sibir-ine: For plugins that a mod author doesn't want cleaning data displayed, we instead omit the cleaning data & leave a comment like `#Do not clean` (see https://github.com/loot/loot.github.io/issues/89 for our rationale & discussion on the topic). However, that's really only for intentional ITMs (which we prefer mod authors don't use in the first place, hence the compromise). AFAICT, deleted references should always be cleaned. I'm not aware of any rationale for not cleaning them as undeleting & disabling has the same effect as deleting without causing the issues that deleted references cause (see this brief description of [deleted references](https://tes5edit.github.io/docs/7-mod-cleaning-and-error-checking.html#DeletedReference) in the xEdit docs). They may not have caused issues in your testing as none of your other mods interacted with those references, but it will cause issues for users with plugins that do. Even if that's extremely unlikely, we would still consider that reason enough for the mod to be cleaned (or, at the very least, not to tell users otherwise).

I can't replace the cleaning data through a suggestion due to the way GitHub handles lines deletions, so you'll have to either replace it manually or remove this cleaning data if you really don't want to submit it yourself (if you remove it, you'll also have to remove *ClockworkTitleRemover.esp*'s entry as a location alone isn't enough metadata to warrant an entry).

masterlist.yaml
```diff
@@ -22290,9 +22317,14 @@ plugins:
     clean:
       - crc: 0xFC69BF72
         util: 'SSEEdit v4.0.3'
+      - crc: 0x3297A568
+        util: 'SSEEdit v4.0.3'
```
sibir-ine: I think this is supposed to be for *ClockworkSorting_Ing_CCCurios_CACO.esp*, which already has this cleaning data.

masterlist.yaml
```diff
@@ -22304,10 +22336,16 @@ plugins:
...
+    msg:
+      - <<: *alreadyInOrFixedByX
+        subs: [ '"Additional Clockwork - Superior Sorting" (ClockworkSuperiorSorting.esp)' ]
+        condition: 'active("ClockworkSuperiorSorting.esp")'
     clean:
       - crc: 0x3297A568
         util: 'SSEEdit v4.0.3'
-  # Clockwork High poly Project Patches #
+      - crc: 0x30BAFDF7
+        util: 'SSEEdit v4.0.3'
```
sibir-ine: I get a different checksum, is this a future version of the plugin?

sibir-ine (changes requested):
Some required changes & it's good to go.

masterlist.yaml
BellCubeDev: Got it.

masterlist.yaml
BellCubeDev: I got that formatting from another plugin (it only had PapyrusExtender if I remember correctly, so it might be different for just one mod), so that might be worth reviewing as well.

sibir-ine: If it's just one item, it can (& we prefer it to) be in-lined (e.g., `req: [ *PapyrusExtender ]`).

masterlist.yaml
BellCubeDev: Dang, you guys are _thurough_.
Nope, not for a future verison. It seems I must've been too tired to discern between the different sorting patches last night—this one's for ClockworkSorting_Ing_CCCurios (already in the masterlist)

masterlist.yaml
BellCubeDev: Yup. My bad.

masterlist.yaml
BellCubeDev: i'm not sure if it works, but I'm trying out github.dev on mobile (away from my desktop ATM). Bit of a pain to use but it's better than the main site crashing my mobile browser almost every time.
Let's see, shall we?

masterlist.yaml
```diff
@@ -15906,24 +15924,26 @@ plugins:
...
+        util: 'SSEEdit v4.0.4'
+  - name: 'ClockworkTitleRemover.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/47087/' ]
-    clean:
-      - crc: 0xCAE2B6AE
-        util: 'SSEEdit v4.0.3g'
+    dirty:
+      - <<: *quickClean
+        crc: 0xBF1FBE1F
+        udr: 3
```
sibir-ine: Missing a `util`, & I get a different checksum.

```suggestion
    dirty:
      - <<: *quickClean
        crc: 0x913191E9
        util: '[SSEEdit v4.0.4](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        udr: 3
```

masterlist.yaml
```diff
@@ -15906,24 +15924,26 @@ plugins:
...
 
   - name: 'Darkend.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/10423/' ]
     dirty:
       # version: 1.4
       - <<: *quickClean
         crc: 0x616730B8
-        util: '[SSEEdit v4.0.0](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 26
```
sibir-ine: This line was unintentionally deleted.

```suggestion
        util: '[SSEEdit v4.0.0](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        itm: 26
```

sibir-ine (changes requested):
Almost ready, just a few more changes required.

masterlist.yaml
BellCubeDev: Aaah. Must've hit cut instead of copy.

sibir-ine (approved):

sibir-ine:
Thanks!

PR #2205, opened 2021-11-24
Update Immersive College of Winterhold

Sabre142:


masterlist.yaml
```diff
@@ -17695,35 +17697,43 @@ plugins:
...
       - Keywords
-      - Names
+      - NPC.AttackRace
       - NPC.Class
+      - NPC.DefaultOutfit
+      - ObjectBounds
       - Outfits.Add
       - Stats
       - Text
-    dirty:
```
BellCubeDev, line 17722: From my (honestly quite limited) understanding, it doesn't seem like removing old CRC entries is a good idea. If a user chooses not to update (or misses it entirely), the dirty flag would, all of a sudden, no longer appear.

Just something for consideration.

sibir-ine: Keeping old cleaning data leads to clutter with diminishing utility, so we tend to just replace it unless users have reason to continue to use an older version (e.g., the author officially supports it).

Sabre142: Pre-Version 6.2 downloads of the mod are not available anymore anyway, and it has been that way for almost a year.

BellCubeDev: > Keeping old cleaning data leads to clutter with diminishing utility, so we tend to just replace it unless users have reason to continue to use an older version (e.g., the author officially supports it).

Cool, got it. I'll keep that in mind.

masterlist.yaml
```diff
@@ -17695,35 +17697,43 @@ plugins:
           - 'Obscure''s College of Winterhold'
           - '[Ultimate College of Winterhold](https://www.nexusmods.com/skyrimspecialedition/mods/44797/)'
         condition: 'active("OCW_Obscure''s_CollegeofWinterhold.esp") and not active("UltimateCollege.esp")'
+      - <<: *patch3rdParty
+        subs:
+          - 'The Gray Cowl Returns! (Creation Club)'
+          - '[Immersive College of Winterhold - More Patches](https://www.nexusmods.com/skyrimspecialedition/mods/28687/)'
+        condition: 'active("ccbgssse020-graycowl.esl") and not active("ICOW_GrayCowlCC_Patch.esp")'
```
sibir-ine: These `*patch3rdParty` messages should be sorted above the other messages.

masterlist.yaml
```diff
@@ -17695,35 +17697,43 @@ plugins:
...
       - Factions
       - Graphics
       - Invent.Add
       - Invent.Remove
       - Keywords
-      - Names
+      - NPC.AttackRace
       - NPC.Class
+      - NPC.DefaultOutfit
+      - ObjectBounds
```
sibir-ine: The object bound edits appear to be CK noise & likely don't serve a purpose in-game.

```suggestion
```

sibir-ine (changes requested):

sibir-ine (approved):

PR #2235, opened 2021-12-11
Add Nightingale Hall Restored - Karliah as a Follower

Sabre142:


masterlist.yaml
```diff
@@ -22097,6 +22097,35 @@ plugins:
         util: '[SSEEdit v4.0.2j](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         nav: 3
 
+  - name: 'Nightingale Hall.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/59324/'
+        name: 'Nightingale Hall Restored - Karliah as a Follower'
+    msg:
+      - <<: *patchProvided
```
sibir-ine: Since the patch is an alternate version of the plugin as opposed to a separate patch plugin, `&useVersion` is more appropriate:

> Use %1% version.

```suggestion
      - <<: *useVersion
```

sibir-ine (changes requested):

sibir-ine (approved):

PR #2260, opened 2021-12-22
v0.15 - SCOS v3.0.0

markekraus:
SCOS v3.0.0 needs to be loaded after Immersive Citizens AI Overhaul due to navmesh changes near Battle-Born Farm.


sibir-ine:
v0.15 is an older version of the masterlist that will no longer be maintained, so you don't need to worry about making pull requests to it, just the current branch (v0.17).

PR #2265, opened 2021-12-31
Update Sentient Chairs of Skyrim

markekraus:
* Add new clean CRC for SCOS 4.1.0 
* version 4.1.0 of SCOS requires SKSE.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -22683,7 +22683,5 @@ plugins:
       - <<: *SKSE
         condition: 'version("MarkekrausSentientChairsOfSkyrim.esp", "4.1.0", >=)'
     clean:
-      - crc: 0xBD12E363
-        util: 'SSEEdit v4.0.4'
       - crc: 0x64014310
         util: 'SSEEdit v4.0.4'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -22679,6 +22679,9 @@ plugins:
     after:
       - 'Immersive Citizens - AI Overhaul.esp'
       - 'Immersive Citizens - OCS patch.esp'
+    req: [ *SKSE ]
     clean:
       - crc: 0xBD12E363
+        util: 'SSEEdit v4.0.4'
+      - crc: 0x64014310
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -22679,6 +22679,9 @@ plugins:
     after:
       - 'Immersive Citizens - AI Overhaul.esp'
       - 'Immersive Citizens - OCS patch.esp'
+    req: [ *SKSE ]
```
sibir-ine: @markekraus You can actually utilize one of LOOT's condition functions & the fact that you put your mod's version number in the plugin's description to reflect that SKSE is only needed for v4.1.0 of your mod. This way, the requirement message will only be displayed if the user doesn't have SKSE installed & they're using v4.1.0 or greater of your mod.

```suggestion
    req:
      - <<: *SKSE
        condition: 'version("MarkekrausSentientChairsOfSkyrim.esp", "4.1.0", >=)'
```

masterlist.yaml
```diff
@@ -22679,6 +22679,9 @@ plugins:
...
     clean:
       - crc: 0xBD12E363
         util: 'SSEEdit v4.0.4'
```
sibir-ine: We usually remove cleaning data for old versions of a mod.

```suggestion
```

sibir-ine (changes requested):

masterlist.yaml
markekraus: awesome... I wasn't sure this would work as I didn't see any examples in the file that does that.. also... this is the first version where the header exists.. will that be a problem?

sibir-ine: Files without version data are interpreted by LOOT as having a version number of 0 for evaluating the version function, so it'll work as intended.

masterlist.yaml
markekraus: Thanks. I was unsure as it's not consistent in the file.

sibir-ine (approved):

sibir-ine:
@markekraus Thanks!

PR #2268, opened 2022-01-02
Update Requiem - The Roleplaying Overhaul

ProbablyManuel:
Requiem already includes all necessary bash tags, but the additional tags added by LOOT cause unexpected behaviour.
- Loot adds `R.AddSpells` on top of  `R.ChangeSpells` from Requiem itself which results in an exception when building a Bashed Patch.
- Loot adds `ObjectBounds`, `Graphics`, and `Sounds`, but Requiem intentionally doesn't add these tags because its edits to them are irrelevant. However, the addition of these tags prevents a Bashed Patch from forwarding relevant edits from other mods.
- Wrye Bash's Merge Leveled List feature doesn't work well with deleveled mods. Requiem even displays a warning if it detects users have created a Bashed Patch with this option enabled. However, the addition of the `Delev` and `Relev` tags prevents the check from working as intended.


sibir-ine (approved):
@Kaenguru-Knobel-Kette Thanks! The location should be changed to point to the new SSE page too. Are there any other locations that need to be updated (e.g., Requiem patches)?

The tags also need to be updated for the LE masterlist.

ProbablyManuel:
The tags in the Oldrim masterlist already match the ones distributed by Requiem v5.0.0.

sibir-ine:
Yeah, but a lot of them are already in LE Requiem's description & can be removed from LOOT.

ProbablyManuel:
The bash tags were added in Requiem v4.0.2. Requiem v3.4.0 still relies on LOOT to get the proper bash tags. There is also other metadata in the Oldrim masterlist that supports v3.4.0, so I don't see the value in removing those tags at this point. Requiem is a slow-moving community 😄

sibir-ine:
Ah, I see. That's fine then.

PR #2271, opened 2022-01-04
Add Chairrows (Chair Arrows)

markekraus:
Add Chairrows mod and set its new SKSE dependency for v2.0.0


sibir-ine (approved):
@markekraus Thanks! Just FYI, the word "Versions" is misspelled in your plugin's description (i.e., "Verions"), but LOOT is still able to recognize "2.0.0" as its version number. Just thought you'd might want to know.

markekraus:
Good catch. I'll fix it on my next update. Thanks!

PR #2272, opened 2022-01-05
Add Chairhammers (Chair Warhammers)

markekraus:
* Add Chairhammers (Chair Warhammers) mod

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -22651,7 +22651,7 @@ plugins:
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/61304/' ]
     req:
       - <<: *SKSE
-        condition: 'version("MarkekrausSentientChairsOfSkyrim.esp", "1.0.0", >=)'
+        condition: 'version("MarkekrausChairhammers.esp", "1.0.0", >=)'
     clean:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -22647,6 +22647,15 @@ plugins:
       - crc: 0x64014310
         util: 'SSEEdit v4.0.4'
 
+  - name: 'MarkekrausChairhammers.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/61304/' ]
+    req:
+      - <<: *SKSE
+        condition: 'version("MarkekrausChairhammers.esp", "1.0.0", >=)'
```
sibir-ine: @markekraus Is v1.0.0 the initial version of the mod, or are there alphas/betas publicly available that don't require SKSE? If so, we shouldn't need to check for the plugin's version number.

markekraus: I had some alphas distributed to some people who were till to do some testing for me. But I guess we can just require it for all versions and I'll just let the beta testers know.

MacSplody (approved):

MacSplody:
@markekraus Thanks.

PR #2282, opened 2022-01-12
Add Cheat Jewelry

markekraus:
Cheat Jewelry was updated from ESL to ESP-FE and now depends on SKSE.


sibir-ine (approved):
@markekraus Thanks!

PR #2303, opened 2022-01-25
Add `compatPatch` message anchor

cecell:
Ref loot/prelude#14

Using pStyl3's suggested wording

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -24,6 +24,12 @@ prelude:
       type: say
       content: 'It is recommended that you read this mod''s [Compatibility Notes](%1%).'
 
+    - &compatPatch
+      type: say
+      content:
+        - lang: en
+          text: 'It is recommended that you check this mod''s [Compatibility Patch Page](%1%).'
+
     - &corrupt
       type: warn
       content: 'This file is corrupt and should not be used.'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -28,7 +28,7 @@ prelude:
       type: say
       content:
         - lang: en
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -24,6 +24,12 @@ prelude:
       type: say
       content: 'It is recommended that you read this mod''s [Compatibility Notes](%1%).'
 
+    - &compatPatch
+      type: say
+      content:
+        - lang: en
+          text: 'It is recommended that you check this mod''s [Compatibility Patch Page](%1%).'
```
sibir-ine: > Also this message could be expanded to mention more than one page of patches should it ever be needed

@cecell, in response to what you [said](https://discord.com/channels/473542112974077963/496196499890241546/935568807697743893) in the Discord, I'd rather the message be as generic as possible from the beginning as changing the message later means voiding any translations provided in the meantime. In this case, it would also mean going back & adjusting each sub where it's used to fit the changed message. It's also worth noting that the page we're directing them to is a compilation of patches as, if it's a page for a single patch or a handful of patches, we'll just add explicit messages instead.

```suggestion
          text: 'It is recommended that you check the following compatibility patch compilation for this mod: %1%'
```

Another thing to note is that you'll also need to add this to the [prelude](https://github.com/loot/prelude) first to actually be able to use it, so please make a PR there too. We can add the example message to the rest of the masterlists, so you don't have to worry about them. However, you're always free to add it to them as well if you want to.

sibir-ine (changes requested):

sibir-ine (approved):
Okay, I'll merge it after it's added to the prelude.

sibir-ine:
@cecell Thanks!

PR #2308, opened 2022-02-02
Add sort to CACO for TechnicolorAlchemy.esp

cecell:


sibir-ine:
CACO's group should result in it loading after by default, & the CACO patches on Technicolor Alchemy Overhaul's page appear to be load-order&ndash;agnostic, so a load after shouldn't be necessary.

cecell:
@sibir-ine LOOT is still sorting CACO (Complete Alchemy) before Technicolor Alchemy.

Here's the 'working version' Technicolor Alchemy that I'm using:
[https://drive.google.com/file/d/1kVQ7DCpTwiUjcQtsmq0peLF3UW_M3M7w/view?usp=sharing](url)

cecell:
May I please resubmit this?

They both directly edit the exact same records in absolutely incompatible ways and CACO's edits should take priority

cecell:
I'm an idiot. If I may, I would like to ask that you please ignore the prior 2 posts and have a great day/week/month/year

PR #2312, opened 2022-02-06
Add Lainalten

cecell:


cecell:
Lainalten heavily edits the area it inhabits and thus must take priority from both a landscape and a navmesh standpoint.  By sorting it after Skyrim at War it solves this issue.  My load order automatically sorted Skyrim at War *after* Lainalten and thus I'm submitting this

sibir-ine:
Is `Skyrim at War.esp` still available legitimately? I found these locations, [here](https://www.nexusmods.com/skyrimspecialedition/mods/49352) & [here](https://bethesda.net/en/mods/skyrim/mod-detail/4213244), but it appears to have been deleted.

**Edit**: I ask because, if it's not, a load after in the masterlist wouldn't be appropriate as we & future maintainers/contributors wouldn't be able to access the plugin to investigate or verify the load after. You can always add a load after to your userlist for personal use.

cecell:
Sure you can! :D All mods are legitimately accessible via the irrevocable license the author has granted the Nexus, and will remain so indefinitely per their TOS: https://www.nexusmods.com/skyrimspecialedition/mods/49352?tab=files&file_id=235075
Nexus TOS: https://help.nexusmods.com/article/18-terms-of-service#RightsYouAre

If you'd like I can add Skyrim at War separately and add the link so there will be a record of its location

pStyl3:
While this PR is about `Lainalten.esp`, since now we also discuss the Skyrim at War mod, I think we will need to discuss how we want to handle links such as the above one for Skyrim at War, or even _if_ we want to use them, before proceeding any further with this PR. The situation here is a bit .. unfortunate.

Skyrim at War has the Nexus modpage ID `49352`. The mod uploaded _after_ that with the ID `49353` is [this](https://www.nexusmods.com/skyrimspecialedition/mods/49353) one and here the original upload date is ` 02 May 2021 12:16PM`. This means that Skyrim at War was uploaded prior to that date.

Around the 20th of June the Nexus implemented it's change so that files no longer could be deleted. This change was unannounced at this point, but some authors quickly [noticed](https://www.reddit.com/r/skyrimmods/comments/o49t79/it_no_longer_seems_to_be_possible_to_delete/) the change. This led to the `Changes to file deletions` topic in the `General Mod Author Discussion` forum on the Nexus (posted on 22nd of June), where the Nexus stuff told mod authors about the change.

And this led to [this](https://www.nexusmods.com/news/14538) news post on the main page of Nexus, posted on the 1st of July. This news reads as follows:
> TL;DR: It is our core belief that making modding easier is a noble goal, so more people can enjoy our joint hobby. We are convinced that our collections system, a project we have spent almost 2 years and many work-hours on, is the way to achieve this goal. For the benefit of this system we have to implement a change to how mod file deletions work. The change was announced to mod authors recently, and basically means that mod files are no longer deleted, but rather archived - which will make them inaccessible unless directly requested e.g. via the API. We understand that not everyone in the mod author community shares our convictions and our vision, and that is why we are granting a 1 month grace period in which mod authors can request to have all their files deleted for good by contacting support@nexusmods.com. 
> 
> 20 July Update: Based on feedback we've received from mod authors who would like to have some but not all of their mod files deleted, we're now offering [partial deletion requests based on this form](https://www.cognitoforms.com/NexusMods/FileDeletionRequestForm), subject to the 5 August deadline.
> 
> (.. Remaining text omitted ..)
___
So Skyrim at War was originally uploaded before the change in Nexus policy was announced, as well as before it was introduced (which as can be seen above happened _before_ the announcement). This means that at the original upload date the mod author of Skyrim at War did not consent with not being able to delete his files, the new system wasn't even introduced or announced at this point.

The important thing now is, that the mod page for Skyrim at War got deleted, and we don't know when exactly that happened.

In cases where mods got uploaded after the 5th of August 2021, where Nexus grace period for mod file deletion ended, and as such each mod upload happened _after_ the new Nexus policy (to not allow file deletions) was introduced - I personally wouldn't be strictly against linking to archived files (even though normal links should be prefered). Uploading a mod to the Nexus after this date means the mod author accepted the new terms (even if they doesn't like the change), and if we deem it necessary to link to an archived mod so that we have the ability to recheck it at a later date, this might be okay.

However, with mods like SaW that got uploaded before the change to Nexus policy, for all we know the mod author could've missed the grace period to properly delete his mod from the Nexus, we simply don't know when exactly the mod page got deleted and what the mod authors intentions were.

And out of that follows that `the irrevocable license the author has granted the Nexus` does not seem like the best description of the situation. And I don't know if it would be good to link to archived files from deleted mod pages, from mods that got uploaded prior to the change, if we don't know the authors intentions here. I think this is not only about what we _can_ do, but what we _should_ do in situations like these.

cecell:
This decision isn't for me to make of course. I can't speak for those that already have it but they aren't a concern.  I'll close it

PR #2322, opened 2022-02-16
Add Thwack - A Creation Club Integration Mod

Sabre142:


sibir-ine (changes requested):
`Gameplay - NPC Balance & Distribution` is more for mods that rebalance or redistribute NPC. Since this mod distributes CC crossbows throughout the game, `Gameplay - Economy & Item Balance` would be more appropriate. It looks good otherwise.

sibir-ine (approved):

PR #2324, opened 2022-02-16
Add Morrowind Threads - A Creation Club Integration Mod

Sabre142:


sibir-ine (changes requested):
`Gameplay - Economy & Item Balance` is more appropriate due to the theme of the mod. It looks good otherwise.

sibir-ine (approved):

PR #2334, opened 2022-02-21
Add Elder Council Amulet - Replacer

Sabre142:


masterlist.yaml
```diff
@@ -2618,6 +2618,15 @@ plugins:
       - crc: 0x92E585C2
         util: 'SSEEdit v4.0.3h (Hotfix 1)'
 
+  - name: 'Praedy''s ElderCouncilAmulet.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/63931/'
+        name: 'Elder Council Amulet - Replacer'
+    msg: [ *optional ]
```
sibir-ine: The optional message is more for plugins that are installed by default but are optional. The mod's FOMOD installer doesn't enable it by default & states that it is optional, so that should suffice:

```suggestion
```

sibir-ine (changes requested):

sibir-ine (approved):

PR #2336, opened 2022-02-21
Update Kthonia's Unique Weapon Pack - Dragonborn Weapons

cecell:


sibir-ine (commented):
You can remove the old cleaning data too, especially since older versions are no longer available.

cecell:
K just removed the old

sibir-ine (approved):
Perfect, thanks!

PR #2339, opened 2022-03-03
Add Bash Tags for some visual replacers

ProbablyManuel:


Sabre142:
- The next release of LOOT will likely expose locations in the UI, so it'd be beneficial to always include mod names for links going forward.

- ESPs with only a single tag can be written in a single line, like this:

```yaml
    tag: [ Graphics ]
```

- It's best to use SSEEdit v4.0.4b for cleaning data, which is available on the xEdit Discord server.

- I'd suggest writing the Unique Uniques entry like this, to have the location and name apply to all of the included ESPs in the mod:

```yaml
  - name: 'Unique Uniques SE.*\.esp'
    url:
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/47725/'
        name: 'Unique Uniques SE Re-Ported'
  - name: 'Unique Uniques SE.esp'
    tag:
      - Graphics
      - ObjectBounds
    clean:
      - crc: 0xC35AC958
        util: 'SSEEdit v4.0.4b'
```

masterlist.yaml
```diff
@@ -15732,6 +15741,15 @@ plugins:
         itm: 18
         udr: 1
 
+  - name: 'StrosMKaiRum.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/43830/'
+        name: 'Unique Stros M''Kai Rum'
+    tag: [ Graphics ]
```
sibir-ine: Since the only non-`Graphics` change is an `Update.esm` forward, it can be tagged with `Deactivate` too. There's also an extra whitespace:

```suggestion
    tag:
      - Deactivate
      - Graphics
```

sibir-ine (changes requested):
@Kaenguru-Knobel-Kette, the `Items` section is more for mods that mainly add new items. Since these mods mainly overhaul the models/textures of existing items, `AudioVisual - Models & Textures` would be more appropriate.

sibir-ine (approved):

PR #2366, opened 2022-04-01
Update Ancient Land

loicreynier:
Add Expand Towns and Cities complete and modular patch


sibir-ine:
The contents of `patchIncluded` are as follows: "You seem to be using **%1%**, but you have not enabled a compatibility patch for this mod. A compatibility patch is included with this plugin's installer." Given the contents, it should only be used on the plugin that contains the patch(es) in its installer, i.e., ETaC in this case. However, both ETaC Complete & ETaC Riverwood already have these messages:

https://github.com/loot/skyrimse/blob/59f948b1d0efc2d196b06a8e94ab263fddec6126/masterlist.yaml#L18895-L18897

https://github.com/loot/skyrimse/blob/59f948b1d0efc2d196b06a8e94ab263fddec6126/masterlist.yaml#L19018-L19020

PR #2375, opened 2022-04-16
Add Woodworker's Whim

BellCubeDev:
New mod, new masterlist entry!

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8820,6 +8820,19 @@ plugins:
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 773
 
+  - name: 'WoodworkersWhim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507'
+        name: 'Nexus Mods'
+    group: *lateChangesGroup
+    clean:
+      # Normal Plugin
+      - crc: 0x24E8B4EA
+        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+      # Translateable Plugin
+      - crc: 0x8CCD892A
+        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+
   - name: 'YASH2.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/2430/' ]
     msg:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8826,10 +8826,23 @@ plugins:
         name: 'Woodworker''s Whim'
     group: *craftingGroup
     after: [ 'Complete Crafting Overhaul_Remastered.esp' ]
+    msg:
+      - <<: *patchProvided
+        subs:
+          - 'Campfire - Complete Camping System'
+        condition: 'active("Campfire.esm") and not active("WoodworkersWhim_Campfire.esp")'
     clean:
       - crc: 0x24E8B4EA
         util: 'SSEEdit v4.0.4b'
       - crc: 0x8CCD892A
+        util: 'SSEEdit v4.0.4b'
+
+  - name: 'WoodworkersWhim_Campfire.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
+        name: 'Woodworker''s Whim'
+    clean:
+      - crc: 0x7F60A661
         util: 'SSEEdit v4.0.4b'
 
   - name: 'YASH2.esp'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8829,7 +8829,7 @@ plugins:
       - crc: 0x24E8B4EA
         util: 'SSEEdit v4.0.4b'
       - crc: 0x8CCD892A
-        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        util: 'SSEEdit v4.0.4b'
 
   - name: 'YASH2.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/2430/' ]
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8822,7 +8822,7 @@ plugins:
 
   - name: 'WoodworkersWhim.esp'
     url:
-      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
         name: 'Woodworker''s Whim'
     group: *lateChangesGroup
     clean:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8835,7 +8835,6 @@ plugins:
         util: 'SSEEdit v4.0.4b'
       - crc: 0x8CCD892A
         util: 'SSEEdit v4.0.4b'
-
   - name: 'WoodworkersWhim_Campfire.esp'
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8828,7 +8828,6 @@ plugins:
     clean:
       - crc: 0x24E8B4EA
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
-      # Translateable Plugin
       - crc: 0x8CCD892A
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 773
 
+  - name: 'WoodworkersWhim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507'
```
sibir-ine: Links should have trailing slashes for consistency & convenience:

```suggestion
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
```

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
+        name: 'Nexus Mods'
```
sibir-ine: The location name should be the name of the mod's page, not just the website it's hosted on. That should only be appended if multiple links with the same name are provided:

```suggestion
        name: 'Woodworker''s Whim'
```

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
+    group: *lateChangesGroup
```
sibir-ine: Can you please elaborate on why you want the plugin placed so late? I'm not saying it shouldn't, I'd just appreciate an explanation of your reasoning.

Checking against my own mod list, I only see conflicts with USSEP, which I assume you've addressed based on your mod's description, & Campfire, which already loads before it since it's an ESM. Are there any specific mods in later-loading groups that it should load after?

From my own experience, the fact that Wrye Bash only preserves the conflict winner for VMAD subrecords is a good reason for it to load later than default, but I'm not sure if that's enough for it to load almost dead last.

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
+    clean:
+      # Normal Plugin
```
sibir-ine: The comments can be removed, the relevant information can be communicated in the commit message:

```suggestion
```

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
+      - crc: 0x24E8B4EA
+        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+      # Translateable Plugin
```
```suggestion
```

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
 
+  - name: 'WoodworkersWhim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507'
+        name: 'Nexus Mods'
+    group: *lateChangesGroup
+    clean:
+      # Normal Plugin
+      - crc: 0x24E8B4EA
+        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
```
sibir-ine: Cleaning data should be submitted using either the latest stable or beta release of xEdit, with the beta release being preferred. QAC reports should also be left unedited, especially since, in this case, Markdown isn't supported for clean cleaning data (only dirty):

```suggestion
      - crc: 0x24E8B4EA
        util: 'SSEEdit v4.0.4b'
```

masterlist.yaml
```diff
@@ -8820,6 +8820,19 @@ plugins:
...
+      # Translateable Plugin
+      - crc: 0x8CCD892A
+        util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
```
```suggestion
      - crc: 0x8CCD892A
        util: 'SSEEdit v4.0.4b'
```

sibir-ine (changes requested):

masterlist.yaml
BellCubeDev: My bad on that. I copy-pasted from the one above to get the link, but it looks like I forgot to check the versioning. And it would seem that my efforts were in vain anyway.

masterlist.yaml
BellCubeDev: The mod changes the VMAD entry for `MineOreScript` to `WWWMineOreAddons` (which "forwards" MineOreScript's code, so to speak, by _extending_ it) and `ResourceFurnitureScript` to `WWWResourceFurnitureAddons` (same *extending* magic). Any overwrite to any of the wood chopping block base records, or, more likely, one of the ore base records, will prevent the mod from functioning on that particular object, which is much more detrimental than the effects of overwriting other kinds of edits. As such, I do believe it to be important that the mod loads almost-last to ensure that the changes are kept.

Additionally, it is relatively labor-intensive to create this kind of manual patch. Drag-and-drop is often not enough to cut it, unlike with, say, a Keyword edit. Assuming that Bash forwards my changes, it should be relatively easy for a user to forward any other changes that may need forwarding.

Changes from USSEP are indeed forwarded. I'll have to check out Campfire's changes and, ~~if needed, maybe make~~ that definitely warrants a patch.

masterlist.yaml
BellCubeDev: 😅

masterlist.yaml
sibir-ine: Another thing I forgot to consider with the `Late Fixes & Changes` group is that it loads after `Dynamic Patches`, which contains things like the Bashed Patch & other user-generated patches. `Late Fixes & Changes` should really only be used for mods that absolutely must load after such patches which I don't think applies in this case.

I think the optimal placement would be in `Crafting` since I just checked & [CCOR](https://www.nexusmods.com/skyrimspecialedition/mods/28608) conflicts with a decent chunk of your mod. A patch would obviously be ideal in this case, but loading your mod after it at least prevents it from overwriting half of your mod while still maintaining most of CCOR's features. AFAIK, none of the later-loading groups should contain any mods that would conflict with yours, & we can always revise your mod's position later if necessary.

Note that since both your mod & CCOR have dependents (e.g., patches), you'd also need to add a load after to prevent cyclic interactions:

```suggestion
    group: *craftingGroup
    after: [ 'Complete Crafting Overhaul_Remastered.esp' ]
```

BellCubeDev:
Heads up—I just added metadata for the Campfire patch I put together (the CCOR patch is going to be a fair bit more annoying to make)

masterlist.yaml
```diff
@@ -8820,6 +8820,31 @@ plugins:
...
+  - name: 'WoodworkersWhim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
+        name: 'Woodworker''s Whim'
+    group: *craftingGroup
+    after: [ 'Complete Crafting Overhaul_Remastered.esp' ]
+    msg:
+      - <<: *patchProvided
+        subs:
+          - 'Campfire - Complete Camping System'
```
sibir-ine: Single element maps should be written in flow style:

```suggestion
        subs: [ 'Campfire - Complete Camping System' ]
```

masterlist.yaml
```diff
@@ -8820,6 +8820,31 @@ plugins:
...
+        condition: 'active("Campfire.esm") and not active("WoodworkersWhim_Campfire.esp")'
+    clean:
+      - crc: 0x24E8B4EA
+        util: 'SSEEdit v4.0.4b'
+      - crc: 0x8CCD892A
+        util: 'SSEEdit v4.0.4b'
+
```
sibir-ine: Plugins from the same page shouldn't be separated by an extra newline:

```suggestion
```

masterlist.yaml
```diff
@@ -8820,6 +8820,31 @@ plugins:
         util: '[SSEEdit v4.0.1](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 773
 
+  - name: 'WoodworkersWhim.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
+        name: 'Woodworker''s Whim'
```
sibir-ine: You can combine the shared locations with a regex plugin, it saves a few lines:

```suggestion
  - name: 'WoodworkersWhim(_Campfire)?\.esp'
    url:
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
        name: 'Woodworker''s Whim'
  - name: 'WoodworkersWhim.esp'
```

masterlist.yaml
```diff
@@ -8820,6 +8820,31 @@ plugins:
...
+    clean:
+      - crc: 0x24E8B4EA
+        util: 'SSEEdit v4.0.4b'
+      - crc: 0x8CCD892A
+        util: 'SSEEdit v4.0.4b'
+
+  - name: 'WoodworkersWhim_Campfire.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64507/'
+        name: 'Woodworker''s Whim'
```
sibir-ine: This location wouldn't be necessary with the regex plugin:

```suggestion
```

sibir-ine (approved):
Thanks!

PR #2380, opened 2022-04-22
Update JK's Skyrim

loicreynier:
This currently fixes the load order between [Rob's Bug Fixes - JK's Skyrim](https://www.nexusmods.com/skyrimspecialedition/mods/63379) and the no lights patch of JK's Skyrim.

Rob's plugin must be loaded before any JK's patch, should I add a `after` for all JK's Skyrim patches in the masterlist?

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -19666,6 +19666,10 @@ plugins:
     inc:
       - 'Blues Skyrim.esp'
       - 'ETaC - Complete.esp'
+  - name: 'JKs_Skyrim_No_Lights_Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/6289/' ]
+    after:
+      - 'JKs Skyrim - Rob''s Bug Fixes.esp'
 
   - name: 'JK''s Temple of Talos.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/56971/' ]
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -19912,7 +19912,6 @@ plugins:
     inc:
       - 'Blues Skyrim.esp'
       - 'ETaC - Complete.esp'
-
   - name: 'JKs[ _]Skyrim_.*_Patch\.esp'
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6289/'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -19912,10 +19912,12 @@ plugins:
     inc:
       - 'Blues Skyrim.esp'
       - 'ETaC - Complete.esp'
-  - name: 'JKs_Skyrim_No_Lights_Patch.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -19666,6 +19666,10 @@ plugins:
     inc:
       - 'Blues Skyrim.esp'
       - 'ETaC - Complete.esp'
+  - name: 'JKs_Skyrim_No_Lights_Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/6289/' ]
+    after:
+      - 'JKs Skyrim - Rob''s Bug Fixes.esp'
```
sibir-ine: Looking at the two in xEdit, they don't actually conflict in any major way. Just the MHDT subrecord of the Tamriel worldspace which probably isn't important to either plugin (also which will most likely get overwritten by something else anyway & I'm not sure even does anything) & a water subrecord in __00009349__ that's probably unintentionally reverted by `JKs_Skyrim_No_Lights_Patch.esp` when it loads after.

That being said, I can see it being the case that either plugin may be updated in a way where load order would matter. In that case, you can use the following regex plugin to cover what appears to be most patches. Feel free to add rules for any others that were missed as you see fit. I've also added a location name & converted the single-element file map to flow style. You'll have to manually place it above all of the other patches but just below `JKs Skyrim.esp`:

```yaml
  - name: 'JKs[ _]Skyrim_.*_Patch\.esp'
    url:
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6289/'
        name: 'JK''s Skyrim'
    after: [ 'JKs Skyrim - Rob''s Bug Fixes.esp' ]
```

sibir-ine (changes requested):

loicreynier:
Sorry it took this long. I removed the No Light patch plugin entry and added your regex plugin entry add the top of JK's Skyrim patches section.

masterlist.yaml
```diff
@@ -24397,6 +24397,13 @@ plugins:
         condition: 'active("Falskaar - SkyTEST and Bruma Patch.esp")'
 
   ### JK's Skyrim Patches ###
+  # JK's Skyrim - Rob's Bug Fixes
+  - name: 'JKs[ _]Skyrim_.*_Patch\.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/6289/'
+        name: 'JK''s Skyrim'
+    after: [ 'JKs Skyrim - Rob''s Bug Fixes.esp' ]
```
sibir-ine: This should be grouped with JK's Skyrim since they're from the same page. Also, the new subsection header isn't necessary.

loicreynier:
Is this what you expected?

masterlist.yaml
```diff
@@ -19913,6 +19913,12 @@ plugins:
       - 'Blues Skyrim.esp'
       - 'ETaC - Complete.esp'
```
sibir-ine: Yes, you just need to remove the newline separating them:

```suggestion
```

sibir-ine (approved):
Thanks!

Issue #2335, opened 2022-02-21
Flat World Map Framework


GunthorNC:
One bit of manual sorting I had to do was for 

[Flat World Map Framework](https://www.nexusmods.com/skyrimspecialedition/mods/29932)

It would not display at all unless I manually moved 
FWMF SE Marked Old Map 4k.esp
FWMF SE Marked Old Map 4k NAT fix.esp
FWMF SE Marked Old Map 4k Vivid Weathers Fix.esp

to the very bottom of my load order. I dunno what the proper placement of them is, but moving them down there allowed my world map to display

Thank you Ortham!

MacSplody:
> One bit of manual sorting I had to do was for
> 
> https://www.nexusmods.com/skyrimspecialedition/mods/62705
> 
> It would not display at all unless I manually moved FWMF SE Marked Old Map 4k.esp FWMF SE Marked Old Map 4k NAT fix.esp FWMF SE Marked Old Map 4k Vivid Weathers Fix.esp
> 
> to the very bottom of my load order. I dunno what the proper placement of them is, but moving them down there allowed my world map to display
> 
> Thank you Ortham!

Other paper map mods are being placed in the Worldspace Settings group, so it should be fine there as well.
Even with that grouping or placing it last for that matter, it may still require patching with other mod changing the Worldspace records. Things like map markers, lighting overhauls, water mods, paper maps will modify Worldspace records.

I'm also reviewing the groups in #2200 
It might be best to split paper maps into their own group.

PR #2407, opened 2022-06-01
Added rules for flat map framework

RyanLangston:
Added Flat World Map Framework

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -24106,12 +24106,10 @@ plugins:
     req: [ *FlatMapMarkers ]
 
   - name: 'FWMF for Paper Map by Mirhayasu.esp'
-    group: 'Late Loaders'
-    after:
-      - 'DynDOLOD.esp'
+    group: 'Worldspace Settings'
     url:
       - 'https://www.nexusmods.com/skyrimspecialedition/mods/29932'
-
+      
   ### Realistic Boat Bobbing Patch Hub ###
   - name: 'RBB - .* Patch\.esp'
     url:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -24105,6 +24105,13 @@ plugins:
     inc: [ 'Paper World Map Unofficial Fix.esp' ]
     req: [ *FlatMapMarkers ]
 
+  - name: 'FWMF for Paper Map by Mirhayasu.esp'
+    group: 'Late Loaders'
+    after:
+      - 'DynDOLOD.esp'
+    url:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -24105,6 +24105,13 @@ plugins:
     inc: [ 'Paper World Map Unofficial Fix.esp' ]
     req: [ *FlatMapMarkers ]
 
+  - name: 'FWMF for Paper Map by Mirhayasu.esp'
+    group: 'Late Loaders'
+    after:
+      - 'DynDOLOD.esp'
```
MacSplody: Both of those rules together are just going to cause a bunch of sorting errors.
As I've mentioned in #2335 

> Other paper map mods are being placed in the Worldspace Settings group, so it should be fine there as well.
Even with that grouping or placing it last for that matter, it may still require patching with other mod changing the Worldspace records. Things like map markers, lighting overhauls, water mods, paper maps will modify Worldspace records.

Generally DynDOLOD should be the last thing you do for your load order.
It will carry forward the last value in each record it copies, so it makes no sense to load other files after it.
If your map isn't working, you either have a conflict or your dynamic patches (Bashed patch, DynDOLOD etc., need to be regenerated.
If `FWMF for Paper Map by Mirhayasu.esp` is placed in the Worldspace Settings group then there shouldn't be much that can overwrite it besides other paper maps, map markers and worldspace tweaks.

RyanLangston:
```
- name: 'FWMF for Paper Map by Mirhayasu.esp'
    group: 'Worldspace Settings'
    url:
      - 'https://www.nexusmods.com/skyrimspecialedition/mods/29932'
```
 Would this work as intended and prevent conflicts

masterlist.yaml
```diff
@@ -24105,6 +24105,11 @@ plugins:
     inc: [ 'Paper World Map Unofficial Fix.esp' ]
     req: [ *FlatMapMarkers ]
 
+  - name: 'FWMF for Paper Map by Mirhayasu.esp'
+    group: 'Worldspace Settings'
+    url:
+      - 'https://www.nexusmods.com/skyrimspecialedition/mods/29932'
```
MacSplody: Needs a few format changes.

```suggestion
  - name: 'FWMF for Paper Map by Mirhayasu.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/29932/' ]
    group: *worldSettingsGroup
```

MacSplody (approved):

MacSplody:
Thanks.

PR #2464, opened 2022-06-29
Update Interesting NPCs

sibir-ine:


masterlist.yaml
```diff
@@ -5396,22 +5396,10 @@ plugins:
...
-      - -C.Music
-      - -C.Name
-      - -C.RecordFlags
-      - -C.Regions
-      - -C.Water
-      - Keywords
-      - Invent.Add
-      - -Invent.Remove
+        name: 'Interesting NPCs'
+    inc: [ 'Open Cities Skyrim.esp' ]
```
MacSplody: You'll probably still get complaints from anyone using the patch, despite the warnings.

sibir-ine: True, we could add a `detail` stating that the official patch is unsupported & not guaranteed to solve incompatibilities as stated by the author of the mod. Or do you think it'd be better to just hide the incompatibility message if they have the patch?

MacSplody: Hiding is fine, though the first message could be added to the patch itself.
Something along the lines of this perhaps.

> This patch is unsupported by its author, use at your own risk as it is not guaranteed to solve incompatibilities.

PR #2476, opened 2022-07-01
Add JK's Redbag's Solitude

robertgk2017:
idk how loot sorts all of these by default, but i believe the load order is supposed to be

1. jk's
2. dawn
3. jks - dawn patch
This part should already be good

4. redbag''s
5. jks redbags
6. jks redbags dawn of skyrim patch
the masters for both jks redbags files are properly set to they are good, perhaps a rule needed for RedBag's Solitude.esp

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -25596,3 +25596,34 @@ plugins:
     clean:
       - crc: 0x111DD990
         util: 'SSEEdit v4.0.4b'
+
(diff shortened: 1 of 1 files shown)


sibir-ine:
From testing, LOOT should sort them this way by default.

sibir-ine (approved):
Thanks!

robertgk2017:
hey i didnt have to fix anything this time!! cool cool!

PR #2477, opened 2022-07-03
Add BUVARP SE RE

Sabre142:


sibir-ine (changes requested):
_Gameplay - NPC Behaviour & Dialogue_ would be a more appropriate.

sibir-ine (approved):
Thanks!

PR #2489, opened 2022-07-17
Add Tamrielic Distribution

Sabre142:


sibir-ine (changes requested):
Since the mod focuses on distributing items from CC & other mods, including but not exclusive to NPCs, **Gameplay - Economy & Item Balance** would be more appropriate.

sibir-ine (approved):
Thanks!

PR #2493, opened 2022-07-21
Update Additional Clockwork

BellCubeDev:
Update the Masterlist entry to...
* Reflect Additional Clockwork's 3.0.1 update
* Add a missing plugin
* Move the `URL` key into a regex for Additional Clockwork plugins

As a note for the regex, it _can_ be modified later to accommodate other plugins that start with `Clockwork` (using fancy-schmancy stuff like Negative Lookaheads). I did a quick run-through in VS Code to see if any other plugins would match, and it didn't find any.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -17528,7 +17528,7 @@ plugins:
       - *PapyrusExtender
       - *SKSE
     clean:
-      - crc: 0xD692DD72
+      - crc: 0xCB190694
         util: 'SSEEdit v4.0.4b'
   - name: 'ClockworkTitleRemover.esp'
     dirty:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -17471,13 +17471,15 @@ plugins:
         itm: 122
 
   # Additional Clockwork
+  - name: 'Clockwork.+?\.esp'
```
sibir-ine: The quantifier doesn't need to be made lazy in this case:

```suggestion
  - name: 'Clockwork.+\.esp'
```

sibir-ine (changes requested):
Putting the location in a regex doesn't make it immediately obvious for maintainers that the patches are from the same page, so you can you please group them with the rest of the plugins from your mod? You can move them all to `3RD Party Patches` since it's an assortment of 3rd-party add-ons & patches for another mod.

sibir-ine:
Since all of these plugins are from the same page, they shouldn't be separated by an extra newline. Looks good otherwise.

sibir-ine (approved):
Thanks!

PR #2508, opened 2022-07-30
Add Protect Unique NPCs

ikonomov:
Being grouped in earlyLoadersGroup could minimize potential issues with other NPC mods.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8459,8 +8459,8 @@ plugins:
         name: 'Protect Unique NPCs'
     group: *earlyLoadersGroup
     inc:
-      - 'NPCs Protected Redux.esp'
       - 'NPCs Protected and Uncapped.esp'
+      - 'NPCs Protected Redux.esp'
       - 'Protected NPCs.esp'
       - 'PyP_Legendary.esp'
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -8453,6 +8453,17 @@ plugins:
...
     tag: [ Actors.ACBS ]
 
+  - name: 'ProtectUniqueNPCs.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/49667/'
+        name: 'Protect Unique NPCs'
+    group: *earlyLoadersGroup
+    inc:
+      - 'NPCs Protected Redux.esp'
+      - 'NPCs Protected and Uncapped.esp'
```
sibir-ine: These should be sorted alphabetically:

```suggestion
      - 'NPCs Protected and Uncapped.esp'
      - 'NPCs Protected Redux.esp'
```

sibir-ine (changes requested):

ikonomov:
@sibir-ine  Ah!  I was wondering about that, but since they come from the same mod I thought it would be appropriate to list the first one that's used as main for the mod.  Thank you, as always!

ikonomov:
Also both are independent plugins with no dependencies on each other coming from the same mod so I thought it would be a good idea to list both.

sibir-ine (approved):
Thanks!

PR #2509, opened 2022-08-05
Update Zim's Immersive Artifacts

DomT602:
Patch no longer required. 
From LOTD page;
Removed : The Zim's Complete conflict resolution patch (No longer needed)


sibir-ine:
@DomT602, the patch can also be removed from here:
https://github.com/loot/skyrimse/blob/02df493725250fa8f48e00b142cd4f83f3b8767d/masterlist.yaml#L24386-L24387

sibir-ine (approved):
Thanks!

PR #2599, opened 2022-10-23
Update Better Dynamic Snow

FelesNoctis:
Shader Tools Updated (SDU) uses the same DLL name and can be substituted into Parallax Shader Fix (PSF), but does not come with the shader files. The snow shaders are the actual conflict with Better Dynamic Snow 3 (BDS).

This updates BDS3's conditionals to specifically check for the shader files so the version warning doesn't appear when a user has SDU installed for other purposes.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2038,12 +2038,14 @@ plugins:
     inc:
       - name: 'SKSE/Plugins/ShaderTools.dll'
         display: 'SSE Parallax Shader Fix'
+        condition: 'file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
       - name: 'SKSE/Plugins/SSEShaderTools.dll'
         display: 'SSE Parallax Shader Fix AE'
+        condition: 'file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
     msg:
       - <<: *useVersion
         subs: [ '2.11' ]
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -2038,12 +2038,14 @@ plugins:
...
         display: 'SSE Parallax Shader Fix'
+        condition: 'file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
       - name: 'SKSE/Plugins/SSEShaderTools.dll'
         display: 'SSE Parallax Shader Fix AE'
+        condition: 'file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
     msg:
       - <<: *useVersion
         subs: [ '2.11' ]
-        condition: 'file("skse/plugins/(SSE)?ShaderTools\.dll")'
+        condition: 'file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
```
sibir-ine: Presumably, the shaders aren't loaded without the plugins, so I'd keep the check for them:
```suggestion
        condition: 'file("skse/plugins/(SSE)?ShaderTools\.dll") and file("skse/plugins/shaders/03[\dA-F]{6}_Parallax_Vc\w*_Snow(_Aam)?\.ps\.hlsl")'
```

FelesNoctis: Yup, probably a good idea.

sibir-ine (approved):
Thanks!

PR #2614, opened 2022-11-08
Add Dawn of RedBag's Solitude

robertgk2017:
Added Czasiors Dawn of RedBag's Solitude and my fixes page for it as well as metadata for both.


robertgk2017:
For sanity sake i tested the download of both files to check the CRC's!

sibir-ine (approved):
Thanks!

PR #2621, opened 2022-11-15
Add Solitude Docks

robertgk2017:
Added entry for Solitude Docks.esp, links and meta data to original mod page, someone else's replacer hotfix, and my up to date rob's bug fixes.

Also a ninja fix to add the .esp extension to my Dawn of RedBag's Solitude entry.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2614,6 +2614,26 @@ plugins:
       - crc: 0x65625CA4
         util: 'SSEEdit v4.0.4b'
 
+  - name: '(man_clavMask|Clavicus mask Reliquary of Myth Patch|Masque of Trickery|REQ_ClavicusMask_patch|Zims clavicus patch)\.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/60764/'
+        name: 'Masque of Clavicus Vile'
+  - name: 'man_clavMask.esp'
+    tag:
+      - Graphics
+      - ObjectBounds
+    clean:
+      - crc: 0x4A090A1F
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x89FE574A
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x71224803
+        util: 'SSEEdit v4.0.4c'
+  - name: 'Masque of Trickery.esp'
+    clean:
+      - crc: 0x0AD24D8F
+        util: 'SSEEdit v4.0.4c'
+
   - name: 'man_DaedricShrines(_Wintersun_Patch)?\.esp'
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/78772/'


robertgk2017:
OOF failed on this hang on ill fix

robertgk2017:
I don't get what's wrong with 26779, i copied that line from an existing mod and just updated the xEdit version letter.

This should be good for the checks.

masterlist.yaml
```diff
@@ -26745,7 +26745,7 @@ plugins:
       - crc: 0x28033725
         util: 'SSEEdit v4.0.4c'
 
-  - name: 'Dawn of RedBag''s Solitude'
+  - name: 'Dawn of RedBag''s Solitude.esp'
```
sibir-ine: I'll correct this in a separate commit as I'd rather keep this PR focused:
```suggestion
  - name: 'Dawn of RedBag''s Solitude'
```

masterlist.yaml
```diff
@@ -26760,3 +26760,33 @@ plugins:
         util: 'SSEEdit v4.0.4c'
       - crc: 0x76F44BFE
         util: 'SSEEdit v4.0.4c'
+        
+  - name: 'Solitude Docks.esp'
```
sibir-ine: There's some extra whitespace here:
```suggestion
        util: 'SSEEdit v4.0.4c'

  - name: 'Solitude Docks.esp'
```

masterlist.yaml
```diff
@@ -26760,3 +26760,33 @@ plugins:
...
+        util: '[SSEEdit v4.0.4c](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 3
+    clean:
+      - crc: 0xA100B5ED
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x5A146CCB
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x6574C43C
+        util: 'SSEEdit v4.0.4c'
+
```
sibir-ine: This extra newline is unnecessary:
```suggestion
```

masterlist.yaml
```diff
@@ -26760,3 +26760,33 @@ plugins:
...
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/13922/'
+        name: 'Solitude Docks'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/33777/'
+        name: 'Solitude Docks Updated'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64071/'
+        name: 'Rob''s Bug Fixes - Solitude Docks Updated'
+    msg:
+      - <<: *patchUpdateAvailable
+        subs: [ '[Rob''s Bug Fixes - Solitude Docks Updated](https://www.nexusmods.com/skyrimspecialedition/mods/64071/)' ]
+        condition: '(checksum("Solitude Docks.esp", 650E43BB) or checksum("Solitude Docks.esp", A100B5ED) or checksum("Solitude Docks.esp", DDAD84FE) or checksum("Solitude Docks.esp", 5A146CCB))'
```
sibir-ine: The extra parentheses are unnecessary:
```suggestion
        condition: 'checksum("Solitude Docks.esp", 650E43BB) or checksum("Solitude Docks.esp", A100B5ED) or checksum("Solitude Docks.esp", DDAD84FE) or checksum("Solitude Docks.esp", 5A146CCB)'
```

masterlist.yaml
```diff
@@ -26760,3 +26760,33 @@ plugins:
...
+        crc: 0xDDAD84FE
+        util: '[SSEEdit v4.0.4c](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 3
+    clean:
+      - crc: 0xA100B5ED
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x5A146CCB
+        util: 'SSEEdit v4.0.4c'
+      - crc: 0x6574C43C
+        util: 'SSEEdit v4.0.4c'
```
sibir-ine: When submitting cleaning data, especially for different versions of a plugin, could you please indicate which checksums correspond to which versions, including version numbers? You can either do so in the commit/PR message or by adding comments labeling each checksum (though don't include version numbers in those, leave those in the commit/PR message). From checking myself, I can see that `0x650E43BB` is the original v1.45, `0xDDAD84FE` is the other replacer v2.1, & `0x6574C43C` is your replacer v2.0.

However, I don't know what the other checksums are. I'm guessing they're the original & the other replacer after QAC (though I get different checksums due to different cleaning settings presumably). If so, they should be removed from the PR. They're not really worth adding since they depend on your cleaning settings & users will know the plugin is clean since they cleaned it themselves. They should be removed from the `patchUpdateAvailable` message condition as well.

sibir-ine (changes requested):

masterlist.yaml
robertgk2017: I always put these in the order of the URL links.

the top most is the original Solitude docks
next is featherstones updated
last is my fixes.

Mine ofc doesn't have a dirty section because i qac it right before upload (also to get the last crc)


thats odd that youd get different crcs.  the crc calculation that xEdit performs hasn't changed in Todd knows how long.  And as far as i know there aren't any cleaning settings that could be different.  QAC is the only cleaning method that i ever user.  Manual cleaning, according to some hehe highly frustrated commentary from Elm, is very much to not be used.

any crcs i input from the original and updated, were from a fresh instal and immediate qac without anything else done to the plugins.

sibir-ine: > I always put these in the order of the URL links.

Right, but this wouldn't work in the case of, e.g., the plugin from the first link being clean & the plugin from second being dirty (the `dirty` key should come before `clean` for stylistic consistency), so it'd be helpful to explicitly list out which checksums correspond to which versions & version numbers regardless going forward.
> And as far as i know there aren't any cleaning settings that could be different.

They're in the **Cleaning** menu in the options:

image
However, they're only for UDRs, so I'm not actually sure why we ended up with different checksums after QAC. Either way, it's all the more reason not to add clean checksums for dirty mods.

masterlist.yaml
```diff
@@ -26760,3 +26760,28 @@ plugins:
...
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/13922/'
+        name: 'Solitude Docks'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/33777/'
+        name: 'Solitude Docks Updated'
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/64071/'
+        name: 'Rob''s Bug Fixes - Solitude Docks Updated'
+    msg:
+      - <<: *patchUpdateAvailable
+        subs: [ '[Rob''s Bug Fixes - Solitude Docks Updated](https://www.nexusmods.com/skyrimspecialedition/mods/64071/)' ]
+        condition: 'checksum("Solitude Docks.esp", 650E43BB) or checksum("Solitude Docks.esp", A100B5ED) or checksum("Solitude Docks.esp", DDAD84FE) or checksum("Solitude Docks.esp", 5A146CCB)'
```
sibir-ine: The clean checksums for the dirty mods should be removed since they're unreliable anyway. You could avoid checksums altogether if you add a version number to your plugin's description. That could be checked instead with the `version()` function:
```suggestion
        condition: 'checksum("Solitude Docks.esp", 650E43BB) or checksum("Solitude Docks.esp", DDAD84FE)'
```

sibir-ine:
In addition to the above, conflicts need to be resolved. Then, it'll be good to go. If the website won't let you resolve them, I can help you do it in Git or do it for you, up to you.

robertgk2017:
You can fix the conflict if its something your familiar with.

What is the nature of the conflict? and is it straightforward to fix?

PR #2622, opened 2022-11-15
Add Masque of Clavicus Vile

Sabre142:


masterlist.yaml
```diff
@@ -2414,6 +2414,26 @@ plugins:
       - crc: 0x796B5447
         util: 'SSEEdit v4.0.3h (Hotfix 1)'
 
+  - name: '(Clavicus mask Reliquary of Myth Patch|man_clavMask|Masque of Trickery|REQ_ClavicusMask_patch|Zims clavicus patch)\.esp'
```
sibir-ine: If the main plugin is `man_clavMask.esp`, I'd put it first in a regex like this & sort by it rather than one of its patches. If the author adds more patches that rank higher alphabetically, you'd have to move the whole group of plugins around to stay consistent:
```suggestion
  - name: '(man_clavMask|Clavicus mask Reliquary of Myth Patch|Masque of Trickery|REQ_ClavicusMask_patch|Zims clavicus patch)\.esp'
```

sibir-ine (changes requested):

sibir-ine:
The plugins just need to be sorted alphabetically within the section now (i.e., between `JS Unique Utopia SE - Daggers - Johnskyrim.esp` & `man_DaedricShrines(_Wintersun_Patch)?\.esp`).

sibir-ine (approved):
Thanks!

PR #2707, opened 2023-02-09
Update Ryn's Whiterun City Limits

loicreynier:
Extend the condition for Ryn's Whiterun City Limits and CC Tundra Homestead patch to account for the JKs Ryns Whiterun combo patch.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -22793,7 +22793,7 @@ plugins:
         subs:
           - 'CC Tundra Homestead'
           - '[Whiterun Exteriors Patch Collection](https://www.nexusmods.com/skyrimspecialedition/mods/78920/)'
-        condition: 'active("cceejsse001-hstead.esm") and not active("Ryns Whiterun City Limits - (Iggath )?Tundra Homestead Patch\.esp")'
+        condition: 'active("cceejsse001-hstead.esm") and not active("Ryns Whiterun City Limits - (Iggath )?Tundra Homestead Patch\.esp") and not active("JKs Ryns Whiterun Exterior - CC Tundra Homestead patch.esp")'
       - <<: *patch3rdParty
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -22793,7 +22793,7 @@ plugins:
         subs:
           - 'CC Tundra Homestead'
           - '[Whiterun Exteriors Patch Collection](https://www.nexusmods.com/skyrimspecialedition/mods/78920/)'
-        condition: 'active("cceejsse001-hstead.esm") and not active("Ryns Whiterun City Limits - (Iggath )?Tundra Homestead Patch\.esp")'
+        condition: 'active("cceejsse001-hstead.esm") and (not active("Ryns Whiterun City Limits - (Iggath )?Tundra Homestead Patch\.esp") or not active("JKs Ryns Whiterun Exterior - CC Tundra Homestead patch.esp"))'
```
sibir-ine, line 22796: @loicreynier, this condition would result in the message only being hidden if one of the patches in the regex ***&*** the JK combo patch are active (or if the CC mod isn't active of course), which presumably isn't what you intended. The following condition would hide it if at least one of any of the three patches is active:
```suggestion
        condition: 'active("cceejsse001-hstead.esm") and not active("Ryns Whiterun City Limits - (Iggath )?Tundra Homestead Patch\.esp") and not active("JKs Ryns Whiterun Exterior - CC Tundra Homestead patch.esp")'
```

sibir-ine (changes requested):

masterlist.yaml
loicreynier: Oh yes I forgot about the regex, thanks.

sibir-ine (approved):
Thanks!

PR #2726, opened 2023-02-25
Update Improved closefaced helmets

loicreynier:
Add a checksum condition to remove the "missing USSEP patch" message if using [Improved Closefaced Helmets - Fixes](https://www.nexusmods.com/skyrimspecialedition/mods/85514) which forwards edits from USSEP.


sibir-ine (approved):
Thanks! The condition will unfortunately stop working as intended if the author ever updates the plugin, but I'll ask to see if they'd be willing to add version metadata to it that we could use instead.

PR #2727, opened 2023-02-25
Update Improved Closefaced Helmets - Fixes

sibir-ine:


loicreynier:
I am confused on how version metadata works in those conditions. The "missing USSEP patch" message is back while using the Fixes plugin after these changes.

sibir-ine:
Update to v1.2 or higher if available, it was added there.

loicreynier:
Oh right I did not understand that the version was parsed from the ESP.

sibir-ine:
Correct, it's parsed from the plugin's description.

PR #2728, opened 2023-02-26
Update Ancient Land

loicreynier:
Add missing patch message for Ryn's and JK's Whiterun exterior mods and Northern Roads.


sibir-ine (changes requested):
Can you please sort these alphabetically by subs amongst the other `patch3rdParty` messages? E.g., *JK's Whiterun Outskirts* should be right after *Enhanced Solitude Docks*. Looks good otherwise.

loicreynier:
Thanks for the review. Messages are now sorted by subs.

sibir-ine (approved):
Thanks!

Issue #2587, opened 2022-10-15
Reevaluate unaliased messages


sibir-ine:
- [x] Children.esp
- [x] ImCh.esm
- [x] Killable Children.esp
- [x] FacegenForKids.esp
- [x] AsharaSkyrimCharacterPresetsReplacer.esp
- [x] Better_Male_Presets.esp
- [x] Kayla_CharPreset.esp
- [x] Lagertha_CharPreset.esp
- [x] Lydia Face Preset 64bit.esp
- [x] Skyrim Immersive Creatures Special Edition.esp
- [x] SAFO - Valdacil.esp
- [x] Andromeda - Unique Standing Stones of Skyrim.esp
- [x] Convenient Horses.esp
- [x] tcbm.esp
- [x] Growl - Werebeasts of Skyrim.esp
- [x] Immersive Jewelry.esp
- [x] Distinct Interiors.esp
- [x] Distinct Interiors - Player Homes.esp
- [x] EEKs Immersive Whiterun.esp
- [x] `(Solitude Skyway SE Estuary Fortress Extension|Solitude Skyway SE Estuary Fortress Access Extension)\.esp`
- [x] TGCoS + Solitude Skyway SE [PATCH - KvHarborside].esp

pStyl3:
Here are the aforementioned messages (I've condensed the relevant parts from the masterlist a bit):

- [X] **Message 1**
`After Character Creation you should deactivate and/or uninstall this.`

<details>
  <summary>Masterlist entries</summary>

```yaml
  - name: 'AsharaSkyrimCharacterPresetsReplacer.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/2406/' ]
    msg:
      - type: say
        content:
          - lang: en
            text: 'After Character Creation you should deactivate and/or uninstall this.'
        condition: 'active("AsharaSkyrimCharacterPresetsReplacer.esp")'

  - name: 'Better_Male_Presets.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/2001/' ]
    msg:
      - type: say
        content:
          - lang: en
            text: 'After Character Creation you should deactivate and/or uninstall this.'
        condition: 'active("Better_Male_Presets.esp")'

  - name: 'Kayla_CharPreset.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/3851/' ]
    msg:
      - type: say
        content:
          - lang: en
[...]

PR #2743, opened 2023-03-11
Update Skyrim Immersive Creatures Special Edition

Sabre142:
Related to #2587.

Let me know if you want to keep the patch3rdParty messages.


sibir-ine (approved):
Thanks!

PR #2804, opened 2023-05-27
Add Torch Brightness Radius

nuiva:
Add [Torch Brightness Radius](https://www.nexusmods.com/skyrimspecialedition/mods/5101).

I copied the *Realistic Torch Light SE.esp* block above since these mods do different flavours of the same thing.

This fixes LOOT placing *Torch Brightness Radius.esp* before popular torch dimmers, such as *ELFX - Hardcore.esp* and *Lux.esp*.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -4887,6 +4887,18 @@ plugins:
       - Graphics
       - Stats
 
+  - name: 'Torch Brightness Radius.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/5101/' ]
+    after:
+      - 'ELE_SSE.esp'
+      - 'ELFX - Hardcore.esp'
+      - 'Lux.esp'
+      - 'Luminosity Lighting Overhaul.esp'
+      - 'RLO - Interiors.esp'
+    tag:
+      - Graphics
+      - Stats
+
   - name: 'RelightingSkyrim_SSE.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/8586/' ]
     dirty:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -4887,18 +4887,6 @@ plugins:
       - Graphics
       - Stats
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -4887,6 +4887,18 @@ plugins:
...
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/5101/' ]
+    after:
+      - 'ELE_SSE.esp'
+      - 'ELFX - Hardcore.esp'
+      - 'Lux.esp'
+      - 'Luminosity Lighting Overhaul.esp'
+      - 'RLO - Interiors.esp'
+    tag:
+      - Graphics
+      - Stats
```
sibir-ine: The **Stats** tag is unnecessary as it only edits the radius & fade value of two light records. Also, you could add the **Deactivate** tag so that the plugin is deactivated when imported into a Bashed Patch, allowing Wrye Bash to import only the relevant changes:

```suggestion
    tag:
      - Deactivate
      - Graphics
```

masterlist.yaml
```diff
@@ -4887,6 +4887,18 @@ plugins:
       - Graphics
       - Stats
 
+  - name: 'Torch Brightness Radius.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/5101/' ]
```
sibir-ine: You can add a name to locations that LOOT can then use when displaying them:

```suggestion
    url:
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/5101/'
        name: 'Torch Brightness Radius'
```

sibir-ine (changes requested):
In addition to the suggestions, please either sort the plugin alphanumerically (ignoring symbols & spaces) within the section or just append it to the end of the masterlist. Otherwise, it looks good to me.

nuiva:
Thanks for the review. I sorted it within the section.

sibir-ine (approved):
Thanks!

PR #2818, opened 2023-06-08
Update Snazzy Interiors Patch Collection

robertgk2017:
Add Windhelm House Clan Shattershield release to the condition.


sibir-ine (approved):
Thanks!

PR #2821, opened 2023-06-09
Update Improved Follower Dialogue - Lydia

SierraKomodo:


masterlist.yaml
```diff
@@ -13931,9 +13931,16 @@ plugins:
...
+    dirty:
+      - <<: *quickClean
+        crc: 0x82652DA7
+        util: '[SSEEdit v4.0.4](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 39
     clean:
       - crc: 0xFFFB3713
         util: 'SSEEdit v4.0.3'
+      - crc: 0x3B8BB25A
+        util: 'SSEEdit v4.0.4'
```
sibir-ine: We don't include the clean checksum for dirty plugins as it can be volatile. The old cleaning data can be removed as well:

```suggestion
```

sibir-ine (changes requested):

masterlist.yaml
SierraKomodo: Updated

sibir-ine (approved):
Thanks!

PR #2830, opened 2023-06-26
Update Snazzy Interiors Patch Collection

robertgk2017:
Add Windhelm Blacksmith and Markarth Treasury House to the list.


sibir-ine (approved):
Thanks!

PR #2837, opened 2023-07-25
Update HS Interior Overhauls

HyliosSykes:
Added new releases of HS Interior Overhauls to the list

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -22516,7 +22516,7 @@ plugins:
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/96227/'
         name: 'HS Windhelm - Temple of Talos'
-  - name: 'HS(DragonBridge|Markarth|PlayerHomes|Riverwood|Solitude|Whiterun|Wilderness|Windhelm) - (Four Shields Tavern|Silver-Blood Inn|Breezehome|Hjerim|Honeyside|Proudspire Manor|SeverinManor|Vlindrel Hall|Alvor and Sigrid''s House|Faendal''s House|Hod and Gerdur''s House|Riverwood Trader|Sleeping Giant Inn|Sven and Hilde''s House|Angeline''s Aromatics|Bits and Pieces|Temple of the Divines|The Winking Skeever|Arcadia''s Cauldron|Belethor''s General Goods|Carlotta Valentia''s House|Dragonsreach|Heimskr''s House|Jorrvaskr|Olava the Feeble''s House|Severio Pelagia''s House|Temple of Kynareth|The Bannered Mare|The Drunken Huntsman|Warmaiden''s|Ysolda''s House|Anise''s Cabin|Sadri''s Used Wares|Temple of Talos)\.esp'
+  - name: 'HS(DragonBridge|Markarth|PlayerHomes|Riverwood|Solitude|Whiterun|Wilderness|Windhelm) - .+\.esp'
     after:
       - 'EnhancedLightsandFX.esp'
       - 'Immersive Citizens - AI Overhaul.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -22468,7 +22508,15 @@ plugins:
...
+  - name: 'HSWindhelm - Sadri''s Used Wares.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/95520/'
+        name: 'HS Windhelm - Sadri''s Used Wares'
+  - name: 'HSWindhelm - Temple of Talos.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/96227/'
+        name: 'HS Windhelm - Temple of Talos'
+  - name: 'HS(DragonBridge|Markarth|PlayerHomes|Riverwood|Solitude|Whiterun|Wilderness|Windhelm) - (Four Shields Tavern|Silver-Blood Inn|Breezehome|Hjerim|Honeyside|Proudspire Manor|SeverinManor|Vlindrel Hall|Alvor and Sigrid''s House|Faendal''s House|Hod and Gerdur''s House|Riverwood Trader|Sleeping Giant Inn|Sven and Hilde''s House|Angeline''s Aromatics|Bits and Pieces|Temple of the Divines|The Winking Skeever|Arcadia''s Cauldron|Belethor''s General Goods|Carlotta Valentia''s House|Dragonsreach|Heimskr''s House|Jorrvaskr|Olava the Feeble''s House|Severio Pelagia''s House|Temple of Kynareth|The Bannered Mare|The Drunken Huntsman|Warmaiden''s|Ysolda''s House|Anise''s Cabin|Sadri''s Used Wares|Temple of Talos)\.esp'
```
sibir-ine, line 22519: Since this regex is getting quite unruly, you could simplify it with something like the following:

```suggestion
  - name: 'HS(DragonBridge|Markarth|PlayerHomes|Riverwood|Solitude|Whiterun|Wilderness|Windhelm) - .+\.esp'
```

The `.` means any character (other than a newline), & the `+` means 1 or more of the previous character, so it's essentially a wildcard. The only potential issue would be if you plan on releasing or have released any mods that would fit this new regex that you don't want the following metadata to apply to. Would that be an issue?

HyliosSykes: I completely agree and I think that's something I wanted to do in the first place (since that would automatically include my new releases without me having to submit new changes to the masterlist).

My only concern would be the patches since I tend to name them like so <Name of the base mod> - <Name of the Patch>.esp (ex : "HSWindhelm - Temple of Talos.esp" would have a patch named "HSWindhelm - Temple of Talos - USSEP.esp")

Those patches will fit the regexp as well but they always contain the base mod as a requirement (in the esp data), would LOOT be able to sort them out properly either way? Patches should come after the base mod in the LO, but now that they have a rule applying to them I'm not sure how they will react with LOOT.

sibir-ine: Yeah, that'll be fine. The rules that get applied to the patches will just be redundant then, they already sort after those plugins since LOOT always sorts plugins after their masters. There isn't any way I can think of to avoid capturing those patches in the regex without you changing your naming convention for them.

HyliosSykes: Let's go with that then, I'll have to do another commit apparently I'm unable to accept your changes above (I get the message "This diff has recently been updated. Refresh and try again." - But refreshing does change anything), unless you can force your Edit on your side.

sibir-ine: Yeah, I think that happens when your PR is out-of-sync with the repo. It'd be easier for you to commit it. You can just push a new commit to this branch, it'll get pushed to the PR as well.

sibir-ine (approved):
Thanks!

PR #2841, opened 2023-08-18
Remove creationClubPlugins message

pStyl3:
@Ortham intends to remove support for the `LOOT` alias in conditions (e.g. `file("LOOT")`). See [this post](https://discord.com/channels/473542112974077963/473543945188802580/1141849093883306158) in Discord.
The expression `file("LOOT")` is used exclusively for the `creationClubPlugins` message.

image

Since this message is referencing very old LOOT versions that likely nobody uses anymore, remove the message anchor entirely.


Ortham:
Not only is the LOOT version referenced very old, it's not compatible with the last few masterlist branches, so anyone on a version that would see the message wouldn't be able to use the current masterlist anyway.

PR #2875, opened 2023-11-11
Update HS Interior Overhauls

HyliosSykes:
Extending the regexp to include my mods from my "Riften" series.


sibir-ine (approved):
Thanks!

PR #2925, opened 2024-04-08
AI Overhaul - NPC Protection relationship update

Xanardi:
Fixed the relationship between protection attribute mods and "AI Overhaul".
If the load order of "AI Overhaul" is later than the mod that changes the protective attribute, the protective attribute will return to its original state.
[AI-Overhaul-Patcher](https://github.com/Excinerus/AI-Overhaul-Patcher/issues/9)

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9889,6 +9889,7 @@ plugins:
   - name: 'Protected NPCs.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/26999/' ]
     group: *earlyLoadersGroup
+    after: [ 'AI Overhaul.esp' ]
     tag: [ Actors.ACBS ]
 
   - name: 'ProtectUniqueNPCs.esp'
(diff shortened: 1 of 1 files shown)


MacSplody:
> Fixed the relationship between protection attribute mods and "AI Overhaul". If the load order of "AI Overhaul" is later than the mod that changes the protective attribute, the protective attribute will return to its original state. [AI-Overhaul-Patcher](https://github.com/Excinerus/AI-Overhaul-Patcher/issues/9)

I don't think this is a good rule to add for a few reasons.
Current groups for those plugins in combination with those load after rules could lead to sorting errors.
And I wouldn't put those back in default group for the masterlist.
Those files were put early to prevent them reverting changes in larger mods.

The user has options to resolve this on their own end.
They can change the group of the plugin through LOOT to default.
This will sort `Protected NPCs.esp` after `AI Overhaul.esp`, if they're both in the same group.

They can use NpcProtector in combination with AI-Overhaul-Patcher.
image

It doesn't patch some NPCs, they think it causes some bugs for those npcs, just so you are aware.
[Removed NPCs that *might* cause bugs ](https://github.com/alsa64/NpcProtector/commit/5cdbd168e56baf62b1ba69492998528e1c481e1d)

Two NPCs didn't receive the flag in testing, results were fine otherwise.
image

sibir-ine:
Closing as the changes have been rejected and the author hasn't responded in several weeks.

PR #2957, opened 2024-07-16
Update Ryn's Farms

loicreynier:
Remove the message for the Skyrim Sewers patch since it is no longer necessary according to the patch page.


sibir-ine (approved):
Thanks!

PR #2975, opened 2024-09-24
Update C.O.I.N. - Coins of Interesting Natures

loicreynier:
Remove patches replaced by the BOS swap and that are now unsupported by the author.


MacSplody (approved):

MacSplody:
Thanks.
The requirements could be updated at some point if there's a reliable way to do it.
Maybe by checking for the `C.O.I.N_SWAP.ini` file.
As we wouldn't want Base Swapper etc. required messages showing up when an old version is used.

PR #2976, opened 2024-09-24
Update Immersive Fort Dawnguard

loicreynier:
Update LotD patch messages. The updated patch from LotD official patches should be used instead of the ones from IFD's FOMod.


MacSplody (approved):

MacSplody:
Thanks.

PR #2978, opened 2024-09-28
Add New Nordic Pots

olmokramer:
I've modeled the changes after the entry for [New Glazed Pots](https://www.nexusmods.com/skyrimspecialedition/mods/69273/) which is a very similar mod by the same author. I verified that the suggested Bash Tags do indeed work as intended and that the mod is clean, both in xEdit 4.1.5f.


MacSplody (approved):

MacSplody:
Thanks.

PR #2993, opened 2024-11-22
fix url mixup of link and name for "Book Covers Skyrim - Lost Library"

nnmrts:
This pull request includes a small change to the `masterlist.yaml` file. The change corrects the order of the `name` and `link` fields for two location entries under the 'Book Covers Skyrim - Lost Library.esp' plugin.


MacSplody:
Thanks.

PR #3000, opened 2024-12-27
Add Cathedral Landscapes - PBR Conversion

sean-kang:
It should overwrite the original Cathedral Landscape plugin.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -8735,6 +8735,14 @@ plugins:
       - crc: 0xA4E835CA
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Cathedral Landscapes PBR.esp'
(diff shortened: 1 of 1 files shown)


sean-kang:
By the way, I am wondering if PBR related plugins should be grouped by its name and automatically placed at the late order. Then pull requests such as mine will be unnecessary.

Sabre142:
Please include the mod's name in the URL and also make the after rule into a single line. It would also be desirable to use the latest xEdit release from the xEdit Discord server for cleaning.

MacSplody:
> By the way, I am wondering if PBR related plugins should be grouped by its name and automatically placed at the late order. Then pull requests such as mine will be unnecessary.

Can't see any reason to group it late at the moment.
Maybe if it had a lot of load after rules.

PR #3021, opened 2025-04-04
KPH was updated, making some ML data obsolete/incorrect

z929669:
Image

That file^ is now obsolete.

To clarify, one now needs two patches when running the latest WACCF and AMB Content Addon instead of the combined patch, since they were parsed from the now-obsolete ESP into two separate, independent patches on KPH. Basically, the message in yellow in the OP should go away when these two plugins exist.

I also presumed some changes to data for [imp_helm_USSEP.esp](https://www.nexusmods.com/skyrimspecialedition/mods/12405).

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -1126,7 +1126,6 @@ plugins:
       - crc: 0xEA917D0B
         util: 'SSEEdit v4.0.4c'
     # v1.6.1170.0
-    clean:
       - crc: 0xEA917D0B
         util: 'SSEEdit v4.1.5d'
   # Saints & Seducers
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -899,15 +899,11 @@ plugins:
         udr: 93
         nav: 3
       # v1.6.1170.0
-      - <<: *reqManualFix
+      - <<: *quickClean
         crc: 0xE6BA88BB
         util: '[SSEEdit v4.1.5d](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 386
         udr: 93
-        nav: 3
-      - <<: *reqManualFix
-        crc: 0xA14EED01
-        util: '[SSEEdit v4.1.5d](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         nav: 3
       # GOG v1.6.659
       - <<: *quickClean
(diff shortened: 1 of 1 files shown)


z929669:
I don't know how to fix:
ERROR: yaml-cpp: error at line 28861, column 7: end of map not found

z929669:
Corrected ERROR: yaml-cpp: error at line 28861, column 7: end of map not found

masterlist.yaml
```diff
@@ -28854,16 +28854,14 @@ plugins:
...
-        subs: [ 'imp_helm_legend_WACCF_AMB_Patch.esp' ]
-        condition: 'active("imp_helm_legend_WACCF_AMB_Patch.esp")'
-      - <<: *useInstead
-        subs: [ 'imp_helm_legend_WACCF_AMB_Patch.esp' ]
-        condition: 'active("aMidianBorn_ContentAddon.esp") and version("aMidianBorn_ContentAddon.esp", "2.0", >=)'
   - name: 'ImprovedClosefacedHelmets_WACCF.esp'
     msg: [ *obsolete ]
+  - name: 'imp_helm_legend_WACCF_AMB_Patch.esp'
+    msg:
+      - <<: *obsolete
```
pStyl3, line 28861: If you write
```yaml
      - <<: *obsolete
```
then either a `subs` or `condition`, or the combination of those two should follow:
```yaml
  - name: 'plugin.esp'
    msg:
      - <<: *obsolete
        subs: [ 'somesub' ]
        condition: 'somecondition'
```
If there's no `subs` or `condition` you want to use, then you could write:
```yaml
    msg:
      - *obsolete
      - <<: *useInstead
        subs: [ 'imp_helm_legend_waccf_patch.esp and imp_helm_legend_amb_patch.esp' ]
        condition: 'active("Weapons Armor Clothing & Clutter Fixes.esp") and active("aMidianBorn_ContentAddon.esp") and version("aMidianBorn_ContentAddon.esp", "3.0", >=)'
```

pStyl3 (changes requested):

masterlist.yaml
z929669: Thank you. I have pushed the change requested with no applicable subs/conditions

pStyl3 (approved):

pStyl3:
Thanks!

PR #3051, opened 2025-04-28
Add Faultier's PBR landscapes

sean-kang:
Better Dynamic Snow and Blended Roads are soft requirements of the mod, and the mod author confirmed that his mod needs to overwrite the two mods.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -9221,6 +9221,17 @@ plugins:
       - crc: 0xFF60C24A
         util: 'SSEEdit v4.0.3'
 
+  - name: 'Faultier''s PBR landscapes.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/125308/' ]
+    after:
+      - 'Better Dynamic Snow SE.esp'
+      - 'BlendedRoads.esp'
+    dirty:
+      - <<: *quickClean
+        crc: 0x9F33E185
+        util: '[TES5VREdit v4.1.5f](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 1
+
   - name: 'gildergreen regrown.esp'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -9221,6 +9221,20 @@ plugins:
...
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/125308/' ]
+    after:
+      - 'Better Dynamic Snow SE.esp'
+      - 'BlendedRoads.esp'
+    dirty:
+      - <<: *quickClean
+        crc: 0x9F33E185
+        util: '[TES5VREdit v4.1.5f](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
+        itm: 1
+    clean:
```
pStyl3, line 9234: We usually don't add `clean` data, if the plugin is `dirty`. As such, please remove the `clean` data.

pStyl3 (changes requested):

masterlist.yaml
sean-kang: I removed the clean data as requested.

pStyl3: Sorry, I should have been more clear. It's not that the whole cleaning report needs to be removed, only the `clean` part of it. So you could do it like this:
```yaml
  - name: 'Faultier''s PBR landscapes.esp'
    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/125308/' ]
    after:
      - 'Better Dynamic Snow SE.esp'
      - 'BlendedRoads.esp'
    dirty:
      - <<: *quickClean
        crc: 0x9F33E185
        util: '[TES5VREdit v4.1.5f](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
        itm: 1
```

sean-kang: Updated again. Thanks.

pStyl3:
Thanks!

PR #3052, opened 2025-04-28
Update checksum for XP32 CTD fix for VR

sean-kang:
The current checksum in the masterlist is from the outdated version of the mod 5 years ago.


pStyl3:
Thanks!

PR #3053, opened 2025-04-28
Update Cathedral Landscape grass only version checksum

sean-kang:
Check the checksum of the latest version of the following grass only variations of the mod.

- Cathedral Landscapes ESP
- Cathedral Landscapes ESPFE
- Cathedral Landscapes - True Grass Only


pStyl3:
Thanks!

PR #3054, opened 2025-04-28
Add Northern Roads

sean-kang:
Northern Roads should load after Vanila Landscape Corrections as the author of VLC recommends.


pStyl3:
Ref:
[Vanilla Landscape Corrections - VLC](https://www.nexusmods.com/skyrimspecialedition/mods/104962/)
[Northern Roads](https://www.nexusmods.com/skyrimspecialedition/mods/77530/)

Screenshot from VLC's modpage:
image

PR #3060, opened 2025-05-01
add DAc0da Paper Map

zyf2008Neptune:


pStyl3:
Is your intention to assign `*paperMapsGroup` to `Sea of Causality Paper Map for FWMF by Limon.esp`? If yes, then instead of adding `DAc0da` to the regex, you'll want to add `Sea of Causality` to it.

pStyl3:
Thanks!

PR #3069, opened 2025-07-01
Add metadata for undo ussep ownership changes.esp

unmasquer:
Adds load metadata for a plugin that reverts most USSEP ownership changes. Ensures it loads after USSEP and is grouped under Fixes, and is not flagged as needing cleaning.


undo ussep ownership changes.esp
pStyl3, line 1: This PR want's to add the mentioned `.esp` to our GitHub repository.
We won't do that, so please remove it from the PR.

pStyl3 (changes requested):

undo ussep ownership changes.esp
unmasquer: That's shockingly dumb and irresponsible, my bad. (Cleary first time using anything like this, should still read tho.) Removing now. Edit: Done.

masterlist.yaml
```diff
@@ -2703,6 +2703,12 @@ plugins:
       - <<: *useBashTweakInstead
         subs: [ 'Tweak Settings -> Msg: No Soul Gem Large Enough' ]
 
+  - name: 'undo ussep ownership changes.esp'
+    url: [ 'https://github.com/unmasquer/Undo-USSEP-Ownership-Changes' ]
```
pStyl3, line 2707: Since we do not add plugins to our GitHub repository, you needed to remove it from the PR. You did, which is good.

However, since you initially uploaded the `.esp` to your *fork of this repository* (which you then based your PR off) & continue to link to your fork - that now is not containing the `.esp` you are linking to - this is a problem. We require a working link for any new plugin entry in the masterlist, and it must fulfill the following:
> The link should refer to an official/canonical/supported source, as opposed to unverified uploads

Can this plugin be found on Nexus? If not, then GitHub may be a possibility, as long as the destination we need to link to is not a fork of this repository. You would need to create your own repository in that case.

unmasquer: Ok, thank you for the clarification and patience.

unmasquer: Confirming, I just need to remove everything from my fork but the masterlist.yaml commit, then link the esp from a repo that isn't a fork of LOOT if I don't want to upload to Nexus? I'll need to modify or resubmit the pull request with the link?

pStyl3: No, the fork in it's current state it alright as is (generally speaking, I may request other changes later), You can see in this very `Files changed` tab of the PR, that only your change to our `masterlist.yaml` is pointing against the original repo, which is exactly what we/you want.

If you were to remove everything else from your fork, this PR would try to do the same to the original repo. Which we don't want.

> then link the esp from a repo that isn't a fork of LOOT

Yes. You need to update the link of this PR, pointing to where the plugin is uploaded.

unmasquer: Okay, I think I got it

unmasquer:
closing. want to make a new one later.

PR #3070, opened 2025-07-02
Changed ParallaxGen.esp to new PG_X.esp name

hakasapl:
Recent versions of PGPatcher (used to be ParallaxGen) use `PG_1.esp`, `PG_2.esp`, etc. where the number can be any number starting from 1 as required to get around the 255 master limit (required for some large load orders).

This PR will change the name to match `PG_X.esp` where `X` is any number.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2832,7 +2832,7 @@ plugins:
     after: [ 'DynDOLOD.esp' ]
     msg: [ *runxLODGen ]
 
-  - name: 'ParallaxGen.esp'
+  - name: 'PG_\d+\.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/120946/' ]
     group: *lateChangesGroup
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -2826,7 +2826,7 @@ plugins:
     after: [ 'DynDOLOD.esp' ]
     msg: [ *runxLODGen ]
 
-  - name: 'ParallaxGen.esp'
+  - name: 'PG_\d+\.esp'
```
MacSplody, line 2829: Should keep the old name as well, for anyone still using that file.
```suggestion
  - name: 'P(arallaxGen|G_\d+)\.esp'
```

masterlist.yaml
```diff
@@ -2832,7 +2832,7 @@ plugins:
...
```
MacSplody, line 2835: Should keep the old name as well, for anyone still using that file.
```suggestion
  - name: 'P(arallaxGen|G_\d+)\.esp'
```

MacSplody:
Thanks.

PR #3077, opened 2025-08-12
Add Happy PBR Trees

sean-kang:
As the author of the mod stated in the mod description, this plugin should overwrite Happy Little Trees plugin. Since there is no dependency between the two plugins, it should be sorted by LOOT.


MacSplody (approved):

MacSplody:
Thanks.

PR #3089, opened 2025-09-28
Update Alternate Perspective

### LINKED ISSUE #3088: Incompability Removal Request: Alternate Start - Alternate Perspective with Realm Of Lorkhan
Alternate Start - Alternate Perspective mod is normally incompatible with realm of lorkhan mod, which is expected.

But there are some patches that makes them both compatible by adding realm of lorkhan as a starting option on alternate perspective mod. So when this patch is installed, these mods are compatible but since LOOT's masterlist has the incompatibility embedded in the Alternate Start's esp, it still gives the incompatilibity pop up on Vortex.

Solution: 
Remove realm of lorkhan from incompatible plugins list for Alternate Perspective

Linext23:
Removing incompatibility line with Alternate Perspective for Realm of Lorkhan

Adding new 3rd party patch message for Alternate Perspective if detected along-side Realm Of Lorkhan

Adding rule to load Alternate Perspective After Realm of Lorkhan

Alternate Perspective - Realm of Lorkhan Patch
https://www.nexusmods.com/skyrimspecialedition/mods/104545

Realm of Lorkhan - Freeform Alternate Start
https://www.nexusmods.com/skyrimspecialedition/mods/18223

Alternate Perspective - Realm of Lorkhan Patch
https://www.nexusmods.com/skyrimspecialedition/mods/50307

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15638,6 +15638,8 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/50307/'
         name: 'Alternate Perspective - Alternate Start'
     group: *altStartGroup
+    after:
+      - 'Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp'
     req:
       - *PapyrusUtil
       - *SilentVoice
@@ -15646,12 +15648,16 @@ plugins:
       - 'Original Opening Scene.esp'
       - 'QuickVanillaStart.esl'
       - 'Random Alternate Start Reborn SE.esp'
-      - 'Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp'
       - 'Skyrim Unbound.esp'
     msg:
       - <<: *patchUnavailable
         subs: [ 'Touring Carriages' ]
         condition: 'active("TouringCarriages.esp")'
+      - <<: *patch3rdParty
+        subs: 
+          - 'Realm of Lorkhan - Freeform Alternate Start'
+          - '[Alternate Perspective - Realm of Lorkhan Patch](https://www.nexusmods.com/skyrimspecialedition/mods/104545/)
+        condition: 'file("Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp") and not file("SHE_LorkhanPerspective.esp")'
     tag:
       - Actors.ACBS
       - Actors.AIData
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6,7 +6,7 @@ prelude:
   # Message Anchors
     - &alreadyInX
       type: error
-      content: 'Delete. Already included in {0}.'  
+      content: 'Delete. Already included in {0}.'
 
     - &alreadyInOrFixedByX
       type: error
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -6,7 +6,7 @@ prelude:
   # Message Anchors
     - &alreadyInX
       type: error
-      content: 'Delete. Already included in {0}.'
+      content: 'Delete. Already included in {0}.'  
 
     - &alreadyInOrFixedByX
       type: error
@@ -15650,14 +15650,14 @@ plugins:
       - 'Random Alternate Start Reborn SE.esp'
       - 'Skyrim Unbound.esp'
     msg:
+      - <<: *patch3rdParty
+        subs:
+          - 'Realm of Lorkhan - Freeform Alternate Start'
+          - '[Alternate Perspective - Realm of Lorkhan Patch](https://www.nexusmods.com/skyrimspecialedition/mods/104545/)'
+        condition: 'file("Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp") and not file("SHE_LorkhanPerspective.esp")'
       - <<: *patchUnavailable
         subs: [ 'Touring Carriages' ]
         condition: 'active("TouringCarriages.esp")'
-      - <<: *patch3rdParty
-        subs: 
-          - 'Realm of Lorkhan - Freeform Alternate Start'
-          - '[Alternate Perspective - Realm of Lorkhan Patch](https://www.nexusmods.com/skyrimspecialedition/mods/104545/)
-        condition: 'file("Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp") and not file("SHE_LorkhanPerspective.esp")'
     tag:
       - Actors.ACBS
       - Actors.AIData
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -15638,8 +15638,6 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/50307/'
         name: 'Alternate Perspective - Alternate Start'
     group: *altStartGroup
-    after:
-      - 'Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp'
     req:
       - *PapyrusUtil
       - *SilentVoice


masterlist.yaml
```diff
@@ -15638,6 +15638,8 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/50307/'
         name: 'Alternate Perspective - Alternate Start'
     group: *altStartGroup
+    after:
+      - 'Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp'
     req:
```
sibir-ine, line 15641:
```suggestion
    group: *altStartGroup
    req:
```
The load after isn't necessary with it's group assignment or with the patch (it resolves all direct conflicts).

masterlist.yaml
```diff
@@ -15646,12 +15648,16 @@ plugins:
...
       - 'Skyrim Unbound.esp'
     msg:
       - <<: *patchUnavailable
         subs: [ 'Touring Carriages' ]
         condition: 'active("TouringCarriages.esp")'
+      - <<: *patch3rdParty
+        subs: 
+          - 'Realm of Lorkhan - Freeform Alternate Start'
+          - '[Alternate Perspective - Realm of Lorkhan Patch](https://www.nexusmods.com/skyrimspecialedition/mods/104545/)
+        condition: 'file("Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp") and not file("SHE_LorkhanPerspective.esp")'
```
sibir-ine, line 15655:
```suggestion
      - <<: *patch3rdParty
        subs:
          - 'Realm of Lorkhan - Freeform Alternate Start'
          - '[Alternate Perspective - Realm of Lorkhan Patch](https://www.nexusmods.com/skyrimspecialedition/mods/104545/)'
        condition: 'file("Realm of Lorkhan - Custom Alternate Start - Choose your own adventure.esp") and not file("SHE_LorkhanPerspective.esp")'
      - <<: *patchUnavailable
        subs: [ 'Touring Carriages' ]
        condition: 'active("TouringCarriages.esp")'
```

There was a missing closing single quote in the second sub causing the CI to fail. Also, this will sort the message aliases alphanumerically & clean up a trailing space.

sibir-ine (changes requested):

masterlist.yaml
Linext23: Realm of Lorkhan does not have the appropriate grouping type (found to be irrelevant after testing, explained below) and is set to default (aka no grouping), if Alternate Perspective doesn't load after Realm of Lorkhan esp, it causes an ILS in-game after clicking new game. 

It's there because Realm of Lorkhan even with being tagged as "Alternate Start" group, it doesn't push Alternate Perspective (And the 3rd party patch by association) after it if it detects it, which is the goal, so unless that rule is added specifically like that, the patch wont properly work in-game out of the box for people who install both with the patch. The collection curator/user could do that but it'd be better to not have them have to do that and requires less extra work.

masterlist.yaml
Linext23: 👍  Nice catch, sorry about that, first time doing this stuff

masterlist.yaml
sibir-ine: The Alternate Start group is sorted after the default group, so, by default, LOOT will sort Alternate Perspective after Realm of Lorkhan as you want. Group rules are soft rules, so they can be overridden by hard rules. I'm assuming you have some combination of custom user rules that is causing LOOT to sort Realm of Lorkhan after Alternative Perspective in your case but not mine.

> And the 3rd party patch by association

This is incorrect; the patch has both mods as masters, so LOOT will always sort it after both mods. These master rules are hard rules & cannot be overridden, instead causing sorting to fail if any user rules are added that would contradict them. If this is not your experience, you're either using a version of the patch that doesn't have Realm of Lorkhan as a master (which doesn't appear to be publicly available for download) or you didn't actually sort with LOOT.

I also tested it & could replicate the infinite loading screen without the patch, but I could not replicate it with the patch. Considering there is no reason for a user to install both mods without the patch (hence the previous incompatibility message), you've now added a message informing them of it, & the patch will always be sorted after both mods, the load rule is purely aesthetic, & we tend to avoid purely aesthetic load rules to provide users with more flexibility. We also only take what is publicly available into consideration, so, if you're using a version of the patch with its masters stripped, a user rule would be more appropriate.

masterlist.yaml
```diff
@@ -6,7 +6,7 @@ prelude:
   # Message Anchors
     - &alreadyInX
       type: error
-      content: 'Delete. Already included in {0}.'
+      content: 'Delete. Already included in {0}.'
```
```suggestion
      content: 'Delete. Already included in {0}.'
```

These trailing spaces were somehow introduced in the last commit.

masterlist.yaml
Linext23: Interesting, the only custom rules i have for ROL is from the collection Immersive and Adult 

Rules for the collection on ROL specifically:
`"pluginRules": {
    "plugins": [
      {
        "name": "realm of lorkhan - custom alternate start - choose your own adventure.esp",
        "after": [
          "nwsfollowerframework.esp",
          "realisticwatertwo.esp"
        ]
      },
      `

So, Nethers Follower Framework (i don't have the water mod because im testing something else) is loading before ROL and ROL is being pushed after it. However, Alternate Perspective still isn't loading after it via Grouping since i cannot find any hard rules that would stop it from loading after via sorting through LOOT. Im not sure how different vortex and the LOOT program sorting are but from my understanding & testing it pulls from the current masterlist when clicking sort.
      
If it doesn't happen without the collection then maybe it's something wrong with the NFF and/or Water rule instead. 
I'll commit the suggested change, just interesting the ILS doesn't happen with the patch when the main Alternate Perspective mod is loaded before Realm of Lorkhan. Seems some investigation is due on our end. Sorry for the misunderstanding, and no, we're using what's all what's available on nexusmods. I'll let the curator (my "boss") know and work with them on the matter. 
      Sorry for wasting time, it's also the first time im really doing any of this so gotta learn somewhere.

masterlist.yaml
Linext23: Sorry 😅 Not even touching that area so im not sure why, should be fixed now

masterlist.yaml
sibir-ine: > Rules for the collection on ROL specifically:
`"pluginRules": { "plugins": [ { "name": "realm of lorkhan - custom alternate start - choose your own adventure.esp", "after": [ "nwsfollowerframework.esp", "realisticwatertwo.esp" ] }, `

Both of those plugins are in groups that sort after the Alternate Start group, so they're making Realm of Lorkhan sort later than Alternate Perspective by proxy. User load after rules are hard rules, so they're causing LOOT to ignore the fact that Realm of Lorkhan's group is typically sorted earlier than NFF's or RW2's groups. LOOT then sorts Alternate Perspective earlier than Realm of Lorkhan because the Alternate Start group is sorted earlier than NFF's & RW2's groups.

> Im not sure how different vortex and the LOOT program sorting are 

They should be the same, I believe Vortex uses libloot, LOOT's API.

> just interesting the ILS doesn't happen with the patch when the main Alternate Perspective mod is loaded before Realm of Lorkhan

The patch resolves all direct conflicts between the two mods, so their relative load order no longer matters with it active.

> Sorry for wasting time, it's also the first time im really doing any of this so gotta learn somewhere.

No worries, thank you for taking the time to submit a PR & discuss your concerns in a civil manner.

sibir-ine (approved):
Thanks.

PR #3109, opened 2026-01-14
add  LuxuryBeachVilla

zyf2008Neptune:


masterlist.yaml
```diff
@@ -18322,6 +18322,7 @@ plugins:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/29932/'
         name: 'Flat World Map Framework'
     group: *paperMapsGroup
+    after: [ '[RR] LuxuryBeachVilla.esp' ]
```
sibir-ine: FWMF is already in the Paper Maps group which loads after most plugins, including the default group. Is there any reason to think a user would move `[RR] LuxuryBeachVilla.esp` to an even later group? If not, this rule isn't necessary.

zyf2008Neptune: ~~LuxuryBeachVilla uses vanilla worldspace data, which overrides FWMF.~~

~~(https://www.loverslab.com/topic/243145-luxury-beach-villa-se/page/4/#findComment-7215786 )~~

zyf2008Neptune: FWMF is loaded after [RR] by default, I'll remove necessary changes.

sibir-ine:
Since the only remaining metadata is a URL, I'll close this PR. They're mostly for maintainers & contributors so that they can verify existing metadata, but no other metadata is being added for this plugin. Feel free to open a new PR if you do have anything else to add.

PR #3116, opened 2026-02-05
Update Protect Unique NPCs

ikonomov:


sibir-ine:
You can add Bash Tags to your plugin's header description, either in Wrye Bash or by adding the line `{{BASH:Actors.ACBS}}` to the plugin's TES4/SNAM field. That way, users will get them even if they don't use LOOT.

ikonomov:
Do you know if they serve any purpose in Vortex or MO2, or just used in Wrye Bash?  Do you think it's not a good idea to add them here?

sibir-ine:
They're only used by Wrye Bash (Mator Smash & xEdit do too, but that's irrelevant to this discussion), but you can use it alongside other mod managers if you're only using it for its utilities like the Bashed Patch. Many users do this, so I do think you should add the tag. Adding Bash Tags via LOOT & via a plugin's description are functionally equivalent (besides precedence, but that's irrelevant here), but, since it's your mod, I wanted to point out that you can add it more directly without needing to use LOOT as a middleman.

ikonomov:
Thanks.  It seems if somebody really cares about their load order because of the type of mods or the number of mods that they use they should use LOOT as the primary source for managing their load order regardless of what tool they use.  It is the one tool that can allow a mod to have its own custom load order rules within a category and in direct relationship to other mods.

sibir-ine:
Right, but that's besides the point. I'm just letting you know that you don't need to add Bash Tags to LOOT when you can add them to your own mod. Wrye Bash grabs tags for plugins from a variety of sources. One of which is your local LOOT masterlist if you have it installed, & another is from the description of the plugin itself. If you really would rather add it via LOOT instead, that's fine, but I want mod authors to be aware that this option exists & that it makes more sense since it's more direct.

ikonomov:
I had the tags added in the plugin descriptions.  I think you directed me to do it a while ago.  I recently noticed that some of the mods in the masterlist have tags, something that I didn't know was possible before, and while doing some updates to my mods yesterday I decided to remove them from the mods themselves and add them here.  I feel like this is a really nice way to have all the load order information in one place.

sibir-ine (approved):

PR #3117, opened 2026-02-05
Update Improved Fog and Lighting

ikonomov:


ikonomov:
Sorry for the barrage of commits.  This is the only other mod that I have which is somewhat problematic in terms of its load order so hopefully these bash tags might be helpful to somebody.  Thank you!

MacSplody (approved):

MacSplody:
Thanks.

PR #3118, opened 2026-02-10
Add Weightless Overhaul Order

hneil9:
Weightless Overhaul has to override Complete Alchemy and Cooking Overhaul to make all the items weightless that has been modified by CA&CO.

Weightless Overhaul: https://www.nexusmods.com/skyrimspecialedition/mods/5407
Complete Alchemy and Cooking Overhaul: https://www.nexusmods.com/skyrimspecialedition/mods/19924

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -30914,3 +30914,9 @@ plugins:
   - name: 'Tonal Architect.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/51199/' ]
     msg: [ *patchUnavailableRequiem ]
+
+  - name: 'WeightlessOV_MERGED.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/5407/'
+        name: 'Weightless Overhaul'
+    after: [ 'Complete Alchemy & Cooking Overhaul.esp' ]


MacSplody:
Not against adding it, but seems like the wrong approach.
Even with the change in order it would still need patching, for direct and indirect conflicts.

There's a newer mod available. [Weightless Anniversary Overhaul - Complete](https://www.nexusmods.com/skyrimspecialedition/mods/82047/)
Which also has patches provided [Weightless Anniversary Overhaul - Patch Pack](https://www.nexusmods.com/skyrimspecialedition/mods/114354/)

hneil9:
Yes I am aware that this might not be perfect but it should be a good start for lots of users since "Weightless Overhaul" is endorsed more than the newer mod "Weightless Anniversary Overhaul - Complete" and is most likely to be installed. For any future direct and indirect conflicts can be addressed later when discovered, but this should give a good starting point.

Also "Weightless Overhaul" gives the option to install components separately without WeightlessOV_MERGED.esp so if a user is having conflicts they should still be able to choose their own order.

Optional Components:
WeightlessOV - Book - BCS.esp
WeightlessOV - Book.esp
WeightlessOV - Ingestible - AP.esp
WeightlessOV - Ingestible.esp
WeightlessOV - Ingredient.esp
WeightlessOV - Misc.esp
WeightlessOV - Scroll.esp
WeightlessOV - SoulGem.esp

PR #3123, opened 2026-02-21
Update Unofficial Skyrim Special Edition Patch

ddbb07:
\* Update cleaning data
\- Version: 4.3.6c

I don't think the do not clean message applies as there are quite a lot of ITMs

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -2309,8 +2309,9 @@ plugins:
         util: '[SSEEdit v4.1.5p](https://www.nexusmods.com/skyrimspecialedition/mods/164/)'
         itm: 78
     clean:
-      - crc: 0xF905CCB4
-        util: 'SSEEdit v4.1.5p'
+      # German
+      - crc: 0xB3AB07FD
+        util: 'SSEEdit v4.0.4'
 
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -2303,11 +2303,14 @@ plugins:
...
+        crc: 0xCE922FDA
+        util: '[SSEEdit v4.1.5p](https://www.nexusmods.com/skyrimspecialedition/mods/164/)'
+        itm: 78
     clean:
-      # Do not clean - 1 ITM that is likely not worth cleaning
-      # German
-      - crc: 0xB3AB07FD
-        util: 'SSEEdit v4.0.4'
+      - crc: 0xF905CCB4
+        util: 'SSEEdit v4.1.5p'
```
sibir-ine: The clean checksum is for a different version. Also, we typically avoid keeping the clean checksums for dirty plugins since it can change depending on your xEdit settings.
```suggestion
      # German
      - crc: 0xB3AB07FD
        util: 'SSEEdit v4.0.4'
```

sibir-ine (changes requested):

masterlist.yaml
ddbb07: My apologies, i did not notice that the clean sum was for a different version

sibir-ine (approved):

PR #3213, opened 2026-04-29
Add group for Standing stone effects

MacSplody:


andycox81-hub:
PR 3213 introduces a LOOT graph branch solely for the Standing Stone Effects group.  Because of this branching structure, if you put any mods into this group, they will sort unpredictably somewhere in between Races and Classes and Core Mods, defeating the point of having this group.  The easiest default structure for users to work with is a straight line.  A simple serial connection, such as Races & Classes, then Standing Stone Effects, then Economy, would avoid this issue while still providing a home for this content category.  Please favor linear integration over branching when adding new groups.

pStyl3:
@andycox81-hub

As per [our documentation](https://loot.readthedocs.io/en/latest/app/usage/groups_editor.html):
> Each group has a name and a list of zero or more other groups it loads after.

In this sense, it is a valid method to create and use LOOT groups, that load after more than one other group - or in other words, it is within scope to create and use branches of groups, and as such, it's also within scope to not strictly adhere to a linear, non-branching connection of all groups.

andycox81-hub:
I understand that it's permissible.  I just don't understand how it's helpful.  Is there some advantage to doing it this way?  Am I understanding the branching system correctly, or have I made a mistake?

PR #3215, opened 2026-06-03
Updated Identity Crisis patching recommendation

AlyssaWatson:
As of [Identity Crisis](https://www.nexusmods.com/skyrimspecialedition/mods/39634) 1.4, a patch for Beyond Skyrim Bruma is no longer necessary nor provided. The author has resolved the conflicts. See pinned post [here](https://www.nexusmods.com/skyrimspecialedition/mods/39634?tab=posts) for details.

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -21044,16 +21044,7 @@ plugins:
         crc: 0xDA11D0DA
         util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 21
-
-  - name: 'BS Bruma Patch.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
-    inc:
-      - name: 'Identity Crisis.esp'
-        display: 'Identity Crisis'
-    msg:
-      - <<: *patchOutdated
-        condition: 'active("BS Bruma Patch.esp")'
-        
+       
   - name: 'Maslea.esm'
     url:
       - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/18327/'
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -21054,7 +21054,17 @@ plugins:
     msg:
       - <<: *patchProvided
         subs: [ 'Beyond Skyrim - Bruma SE' ]
-        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp")'
+        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp") and version("Identity Crisis.esp", <, "1.5")'
+
+  - name: 'BS Bruma Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
+    inc:
+      - name: 'Identity Crisis.esp'
+        display: 'Identity Crisis'
+    msg:
+      - <<: *patchOutdated
+        condition: 'active("BS Bruma Patch.esp") and version("Identity Crisis.esp", >=, "1.5")'
+        
 
   - name: 'Maslea.esm'
     url:
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -21045,17 +21045,6 @@ plugins:
         util: '[SSEEdit v4.0.3](https://www.nexusmods.com/skyrimspecialedition/mods/164)'
         itm: 21
 
-  - name: 'Identity Crisis.esp'
-    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
-    inc:
-      - name: 'BSHeartland.esm'
-        display: 'Beyond Skyrim - Bruma SE'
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -21054,7 +21054,17 @@ plugins:
...
       - <<: *patchProvided
         subs: [ 'Beyond Skyrim - Bruma SE' ]
-        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp")'
+        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp") and version("Identity Crisis.esp", <, "1.5")'
+
+  - name: 'BS Bruma Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
+    inc:
+      - name: 'Identity Crisis.esp'
+        display: 'Identity Crisis'
```
sibir-ine: The message is sufficient, & the incompatibility is inaccurate if they're using an older version of Identity Crisis:
```suggestion
```

masterlist.yaml
```diff
@@ -21054,7 +21054,17 @@ plugins:
     msg:
       - <<: *patchProvided
         subs: [ 'Beyond Skyrim - Bruma SE' ]
-        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp")'
+        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp") and version("Identity Crisis.esp", <, "1.5")'
```
sibir-ine: The `version` function depends on the version read from the plugin's description. The plugin doesn't have any such version in its description, so this won't work. Better to just remove the message as neither v1.4 nor the patch are readily available for download anyway.

masterlist.yaml
```diff
@@ -21054,7 +21054,17 @@ plugins:
...
+
```
sibir-ine: No newlines between plugins from the same page:
```suggestion
```

masterlist.yaml
```diff
@@ -21054,7 +21054,17 @@ plugins:
...
+  - name: 'BS Bruma Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
+    inc:
+      - name: 'Identity Crisis.esp'
+        display: 'Identity Crisis'
+    msg:
+      - <<: *patchOutdated
+        condition: 'active("BS Bruma Patch.esp") and version("Identity Crisis.esp", >=, "1.5")'
+
```
sibir-ine: Excess newline/whitespace:
```suggestion
```

sibir-ine (changes requested):

masterlist.yaml
AlyssaWatson: Sorry, been away traveling since you reviewed. If that's the case, would it be better to remove all compatibility notes for Identity Crisis? I would agree with that statement, especially due to the author's tendency to use very generic naming for their patches. Let me know if this is not preferable.

masterlist.yaml
AlyssaWatson: Thank you for explaining the usage. I'll be sure to use that correctly in the future.

masterlist.yaml
AlyssaWatson: Noted and no longer an issue.

masterlist.yaml
AlyssaWatson: Noted and fixed.

masterlist.yaml
```diff
@@ -21054,7 +21054,17 @@ plugins:
...
+        condition: 'active("BSHeartland.esm") and not active("BS Bruma Patch.esp") and version("Identity Crisis.esp", <, "1.5")'
+
+  - name: 'BS Bruma Patch.esp'
+    url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/39634/' ]
+    inc:
+      - name: 'Identity Crisis.esp'
+        display: 'Identity Crisis'
+    msg:
+      - <<: *patchOutdated
+        condition: 'active("BS Bruma Patch.esp") and version("Identity Crisis.esp", >=, "1.5")'
```
sibir-ine, line 21066: Since `version` doesn't work here as previously stated, you could either use `checksum` or make it unconditional. Checksum would be fine if the mod is unlikely to receive another update any time soon, but making it unconditional would be more reliable, & LOOT supports hiding individual messages. Also, `alreadyInOrFixedByX` would be clearer here, `patchOutdated` doesn't communicate that the patch is no longer necessary. Here's an example:
```yaml
      - <<: *alreadyInOrFixedByX
        type: say
        subs: [ 'Identity Crisis v1.5' ]
```

sibir-ine: Sorry, I meant to comment this with my previous batch of comments, but I guess I forgot to actually post it. Either this or completely removing the entry are fine.

pStyl3: @AlyssaWatson Please take a look at sibir's review once more. If the necessary changes are made, we can merge this PR.

PR #3219, opened 2026-06-18
Add Favor Quests Separated - Headhunter Patch

mick-lue:
This adds an entry for a patch I made, which should load after other plugins editing bounty quests and also has an incompatibility with an optional file from Favor Quests Separated.
There are 2 versions of the patch, but both of them should load after Missives.esp as that is also editing bounty quests.

I didn't know where to put it so I added it to the end of the masterlist.
[x] I did test the change locally.

Reference: https://www.nexusmods.com/skyrimspecialedition/mods/180452

### DIFF
--- a/masterlist.yaml
+++ b/masterlist.yaml
@@ -28407,6 +28407,13 @@ plugins:
       - C.ImageSpace
       - C.Light
 
+  - name: 'Favor Quests Separated - Headhunter( Missives)? Patch\.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/180452/'
+        name: 'Favor Quests Separated - Headhunter Patch'
+    after: [ 'BountiesRedone_MissivesExtension.esp' ]
+    inc: [ 'Favor Quests Seperated - Bounty Quests.esp' ]
+
   - name: 'Fixes for Armor and Clothing Extension.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/28059/' ]
     msg:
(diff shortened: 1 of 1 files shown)


masterlist.yaml
```diff
@@ -30058,3 +30058,12 @@ plugins:
...
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/51199/' ]
     msg: [ *patchUnavailableRequiem ]
+
+  - name: 'Favor Quests Separated - Headhunter( Missives)? Patch\.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/180452'
+        name: 'Favor Quests Separated - Headhunter Patch'
+    after:
+      - 'Missives.esp'
+      - 'BountiesRedone_MissivesExtension.esp'
```
sibir-ine: The `Missives.esp` load rule is unnecessary since the default patch doesn't have any conflicting records, & the Missives patch already has it as a dependency:
```suggestion
    after: [ 'BountiesRedone_MissivesExtension.esp' ]
```

masterlist.yaml
```diff
@@ -30058,3 +30058,12 @@ plugins:
   - name: 'Tonal Architect.esp'
     url: [ 'https://www.nexusmods.com/skyrimspecialedition/mods/51199/' ]
     msg: [ *patchUnavailableRequiem ]
+
+  - name: 'Favor Quests Separated - Headhunter( Missives)? Patch\.esp'
+    url:
+      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/180452'
```
sibir-ine: Just to match the rest of the links:
```suggestion
      - link: 'https://www.nexusmods.com/skyrimspecialedition/mods/180452/'
```

sibir-ine (changes requested):
You can sort it alphabetically within the `3rd Party Patches` section (search for `###### 3rd Party Patches ######`).

masterlist.yaml
mick-lue: oh you are right. I must have misremembered that one. I'll push an update with all your suggested changes in a few minutes.

masterlist.yaml
mick-lue: nexusmods actually removes the slash at the end when visiting the website. But it makes sense for consistency

sibir-ine (approved):

PR #3230, opened 2026-07-23
Update Particle Patch ENB Helper condition for KiLoader distribution

zhuabaobao123:
Some distributions of ENB Helper SE install the DLL to `KiLoader/Plugins/ENBHelperSE.dll` instead of the standard `SKSE/Plugins/` path. One example is the KiLoader distribution on Nexus: https://www.nexusmods.com/skyrimspecialedition/mods/99406

This updates the `Particle Patch` recommendation message condition so it does not appear when ENB Helper is installed via KiLoader.


pStyl3:
Thanks.

